PR #6, opened 2025-08-19
fix: Standardize line endings and update binary file patterns

hyunhee-jo:
Updated the .gitattributes file to enforce LF line endings for all text files. This ensures consistent line endings across different operating systems, preventing potential issues with scripts and development tools. Also, expanded the list of binary file patterns to prevent Git from performing line-ending conversions on them.

**Checklist:**


hnc-leebd:
LGTM

PR #14, opened 2025-09-10
fix: Prevent full path disclosure in JSON output

hnc-hyunheejo:
**Checklist:**


hnc-leebd:
LGTM

PR #26, opened 2025-09-15
ci: pom.xml profile for release

hnc-leebd:
Modified to exclude the ci process that is only needed for distribution, such as gpg signing.

**Checklist:**


hnc-hyunheejo (approved):
LGTM

PR #40, opened 2025-09-19
fix: Modified markdown-with-html to extract more naturally.

hnc-leebd:
- Only table objects have been converted to HTML.
- Each object is separated by a double line break.

**Checklist:**


hnc-hyunheejo (approved):
LGTM

PR #44, opened 2025-09-22
fix: html table to output header row in markdown as well

hnc-leebd:
**Checklist:**


hnc-hyunheejo (approved):
LGTM

PR #45, opened 2025-09-22
fix: Modified markdown-with-html to extract more naturally.

hnc-leebd:
- support list
- modify rules for linebreak
- support table header for markdown-with-html
- change replace char for invalid

**Checklist:**

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/markdown/MarkdownGenerator.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/markdown/MarkdownGenerator.java
@@ -7,8 +7,8 @@
  */
 package org.opendataloader.pdf.markdown;
 
-import org.opendataloader.pdf.containers.StaticLayoutContainers;
 import org.opendataloader.pdf.api.Config;
+import org.opendataloader.pdf.containers.StaticLayoutContainers;
 import org.verapdf.wcag.algorithms.entities.IObject;
 import org.verapdf.wcag.algorithms.entities.SemanticHeading;
 import org.verapdf.wcag.algorithms.entities.SemanticParagraph;
@@ -60,7 +60,11 @@ public class MarkdownGenerator implements Closeable {
         try {
             for (int pageNumber = 0; pageNumber < StaticContainers.getDocument().getNumberOfPages(); pageNumber++) {
                 for (IObject content : contents.get(pageNumber)) {
+                    if (!isSupportedContent(content)) {
+                        continue;
+                    }
                     this.write(content);
+                    writeContentsSeparator();
                 }
             }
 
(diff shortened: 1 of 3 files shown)


MaximPlusov:
@hnc-leebd 
As I understand you're making these changes to improve the statistics here https://github.com/opendataloader-project/opendataloader-dp-bench. Some of the changes are really good and useful, but some of them, in my opinion, probably may make the markdown worse for the end user (for example https://github.com/opendataloader-project/opendataloader-pdf/commit/d5286fcf66225c61e2a402984782fbc725dfc518). Maybe we can add a separate special option for generate markdown for statistic?

hnc-leebd:
Hello! Feedback is always welcome!

We should keep in mind that we are developing a data loader for AI (RAG, LLM, etc.).

For Markdown targeting AI usage, three points are important:

1. The formatting should match the actual rendered result (structured document).
2. The raw format should remain human and AI readable.
3. The rendered result should also be human and AI readable.

Unlike plain text, structured formatting in Markdown helps RAG identify important information, and can be leveraged for chunking, splitting, and many other tasks.

In addition, unlike VLM, LLM generally consumes the raw text format, not the rendered result — which means raw readability is important.

Let's revisit the previous commit.
https://github.com/opendataloader-project/opendataloader-pdf/commit/d5286fcf66225c61e2a402984782fbc725dfc518

**First Change: Preserving Paragraph Boundaries**

There was an issue where paragraph structures inferred during layout analysis were being lost.
When generating Markdown, two paragraphs were exported with only a single line break:

```
paragraph1
paragraph2
```

Markdown treats this as a single paragraph, and the rendered result appears as:

```
paragraph1 paragraph2
```

Remember that markdown is a format where the rendering result matches the actual format!

Paragraph boundaries are important for semantic separation and serve as key anchors for chunking/splitting. Therefore, they must be preserved. so, I modified it accordingly.

**Second Change: Maintaining List Structures**

In the markdown-with-html option, list structures were being converted into semi-structured HTML:

```
- item 1 of first list
- item 2 of first list

item 1 of second list
item 2 of second list
```

was transformed into:

```html
<ul style="list-style-type:none;">
  <li>- item 1 of first list</li>
  <li>- item 2 of first list</li>
</ul>
<ul style="list-style-type:none;">
  <li>item 1 of second list</li>
  <li>item 2 of second list</li>
</ul>
```

For the first list, the rendering looked natural since duplicate markers were avoided.
However, this hides the actual structural markers, making the document semi-structured.

For example, enlarging text and applying bold might visually mimic a heading, but it is not a semantic heading.
Since LLMs consume text rather than rendering, this mismatch can introduce subtle semantic inconsistencies.

This is one reason Markdown is generally preferred over HTML or OOXML (docx) for AI use cases.

In the same vein, when using Markdown for AI, it's much more advantageous to use the Markdown list format rather than the styled HTML list format. Therefore, I modified it accordingly.

Ideally, `bullet markers` should be parsed to distinguish `unordered` vs. `ordered` lists.
While that may be challenging for now, we should at least keep Markdown output both structured and readable in its raw form.

**Third Change: Removing Unnecessary Artifacts**

Elements used only for rendering (e.g., `&nbsp;`) were removed, since they hurt raw readability.

**Summary of Fixes in This Commit**

1. [Raw readability] Avoid redundant line breaks caused by non-rendered elements like `line` or `line-art`.
2. [Raw readability] Use Markdown lists instead of HTML for better structured representation.
3. [Structured formatting] Export table headers using Markdown table syntax.
4. [Structured formatting] For inner table contents where Markdown falls short, allow minimal HTML.

All of these modifications are intended to format Markdown appropriately for AI feeding, and are not intended to improve benchmark results. I wrote this PR based solely on the criteria of whether the markdown format suitable for AI feeding, as explained above, was extracted while testing about 200 documents.

To improve benchmark results, the layout inference performance itself needs to be improved. We need to discuss in other issue how to improve layout inference performance.

MaximPlusov (approved):

PR #89, opened 2025-11-03
Add option for using structure tree

MaximPlusov:
**Checklist:**

### DIFF
--- a/python/opendataloader-pdf/src/opendataloader_pdf/wrapper.py
+++ b/python/opendataloader-pdf/src/opendataloader_pdf/wrapper.py
@@ -248,6 +248,7 @@ def main(argv=None) -> int:
     )
     parser.add_argument(
         "--use-struct-tree",
+        action="store_true",
         help="Enable processing structure tree (disabled by default)",
     )
     args = parser.parse_args(argv)


MaximPlusov:
@codex review

hnc-leebd:
@codex review

MaximPlusov:
@codex review

python/opendataloader-pdf/src/opendataloader_pdf/wrapper.py
```diff
@@ -238,6 +246,10 @@ def main(argv=None) -> int:
         "--replace-invalid-chars",
         help="Replacement character for invalid or unrecognized characters.",
     )
+    parser.add_argument(
+        "--use-struct-tree",
+        help="Enable processing structure tree (disabled by default)",
+    )
```
hnc-leebd, line 253: the new `argparse` flag is declared without `action="store_true"`, so `--use-struct-tree` now requires a value and raises “expected one argument” when users follow the documentation (opendataloader-pdf --use-struct-tree …). As a result the flag can’t be consumed from the Python CLI. Mark it as `action="store_true"` (and keep the default False) so it behaves like the other boolean switches.

hnc-leebd (approved):

PR #123, opened 2025-12-15
Cluster table detection

MaximPlusov:
**Checklist:**

### DIFF
--- a/python/opendataloader-pdf/src/opendataloader_pdf/wrapper.py
+++ b/python/opendataloader-pdf/src/opendataloader_pdf/wrapper.py
@@ -106,7 +106,7 @@ def convert(
     keep_line_breaks: bool = False,
     replace_invalid_chars: Optional[str] = None,
     use_struct_tree: bool = False,
-    table_method: Optional[List[str]] = None,
+    table_method: Optional[Union[str, List[str]]] = None,
     reading_order: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


python/opendataloader-pdf/src/opendataloader_pdf/wrapper.py
```diff
@@ -153,6 +160,9 @@ def convert(
         args.append(replace_invalid_chars)
     if use_struct_tree:
         args.append("--use-struct-tree")
+    if table_method:
+        args.append("--table-method")
+        args.append(table_method)
```
hnc-leebd: Could you please check the syntax once more?

```python
# Example demonstrating the difference between append() and extend()

# Original list
numbers = [1, 2, 3]

# Another list to add
new_values = [4, 5, 6]

# -----------------------------------------
# Using append()
# -----------------------------------------
numbers_append = numbers.copy()

# append() adds the entire object as a single element
numbers_append.append(new_values)

print(numbers_append)
# Output: [1, 2, 3, [4, 5, 6]]

# -----------------------------------------
# Using extend()
# -----------------------------------------
numbers_extend = numbers.copy()

# extend() iterates over the given iterable and adds each element individually
numbers_extend.extend(new_values)

print(numbers_extend)
# Output: [1, 2, 3, 4, 5, 6]
```

PR #141, opened 2025-12-29
Improve paragraph detection

MaximPlusov:


hnc-leebd:
I think it would be good to fix the typo.

Fix typo: areSupportNotSingeLines → areSupportNotSingleLines

PR #164, opened 2026-01-22
feat(hybrid): add formula enrichment support with LaTeX extraction

hnc-leebd:
## Summary

- Add `--enrich-formula` CLI option to hybrid server for LaTeX formula extraction
- Add `--enrich-picture-classes` CLI option for picture classification
- Create dedicated `SemanticFormula` entity for formula output in JSON, Markdown, HTML
- Add unit tests for formula parsing in `DoclingSchemaTransformer`

## Changes

| File | Description |
|------|-------------|
| `SemanticFormula.java` | New entity class for formula elements |
| `FormulaSerializer.java` | JSON serializer with `type: "formula"`, `content: "<latex>"` |
| `DoclingSchemaTransformer.java` | Handle `label: "formula"` from docling |
| `MarkdownGenerator.java` | Output `$$...$$` block syntax |
| `HtmlGenerator.java` | Output `\[...\]` MathJax-compatible syntax |
| `hybrid_server.py` | Add `--enrich-formula`, `--enrich-picture-classes` options |
| `README.md`, `CLAUDE.md` | Documentation updates |

## Output Examples

**JSON:**
```json
{
  "type": "formula",
  "page number": 1,
  "bounding box": [226.2, 144.7, 377.1, 168.7],
  "content": "\\frac{f(x+h) - f(x)}{h}"
}
```

**Markdown:**
```markdown
$$
\frac{f(x+h) - f(x)}{h}
$$
```

**HTML:**
```html
<div class="math-display">\[\frac{f(x+h) - f(x)}{h}\]</div>
```

### Quick test with formula PDF (01030000000143.pdf)

**Terminal 1: Start hybrid server with formula enrichment**
```bash
cd python/opendataloader-pdf && pip install -e ".[hybrid]"
opendataloader-pdf-hybrid --enrich-formula --log-level debug
```

**Terminal 2: Process PDF**
```bash
# Build CLI (if needed)
cd java && mvn package -DskipTests -q

# output
java -jar java/opendataloader-pdf-cli/target/opendataloader-pdf-cli-*-shaded.jar \
  --hybrid docling-fast --hybrid-mode full --format json,markdown,html \
  tests/benchmark/pdfs/01030000000143.pdf
```

### Unit tests
```bash
cd java && mvn test -Dtest=DoclingSchemaTransformerTest
```

- [ ] Hybrid server starts with `--enrich-formula` option
- [ ] JSON output contains `type: "formula"` elements with LaTeX in `content` field
- [ ] Markdown output contains `$$...$$` blocks
- [ ] HTML output contains `<div class="math-display">\[...\]</div>`
- [ ] Unit tests pass

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,6 +58,32 @@ opendataloader-pdf-hybrid --ocr-lang "ja" --force-ocr  # Japanese with forced OC
 - `--ocr-lang`: Comma-separated [EasyOCR language codes](https://www.jaided.ai/easyocr/) (default: EasyOCR default languages)
 - `--force-ocr`: Force full-page OCR on all pages regardless of embedded text
 
+**Enrichment Options:**
+```bash
+# With formula enrichment (LaTeX extraction)
+opendataloader-pdf-hybrid --enrich-formula
+
+# With picture classification
+opendataloader-pdf-hybrid --enrich-picture-classes
+
+# Combined: OCR + enrichments
+opendataloader-pdf-hybrid --ocr-lang "en" --enrich-formula --enrich-picture-classes
+```
+
+- `--enrich-formula / --no-enrich-formula`: Enable formula recognition (extracts LaTeX representation)
+- `--enrich-picture-classes / --no-enrich-picture-classes`: Enable picture classification (bar_chart, pie_chart, flow_chart, etc.)
+
+**Formula Processing Note:**
+When using `--enrich-formula`, use `--hybrid-mode full` on the client side to route all pages to the backend:
+```bash
+# Server
+opendataloader-pdf-hybrid --enrich-formula
+
+# Client
+opendataloader-pdf --hybrid docling-fast --hybrid-mode full input.pdf
+```
+This is required because the formula enrichment model runs on the backend, not in Java processing.
+
 ## Architecture
 
 ```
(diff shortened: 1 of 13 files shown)


hnc-leebd:
Thanks for the thorough review! I've analyzed each issue you raised:

### 1. Missing Line Break After Formula in Markdown

> The `writeFormula` method doesn't add a line break after the closing `$$`

This is already handled. Looking at `writeToMarkdown()` (line 64-65):

```java
this.write(content);
writeContentsSeparator();  // Adds two line breaks after every content
```

Every content element (including formulas) gets `writeContentsSeparator()` called after it, which writes two line breaks. This is the same pattern used for all other content types - none of them add trailing line breaks in their individual write methods.

### 2. Potential HTML Escaping Issue

> `getCorrectString()` may escape LaTeX backslashes incorrectly

Checked the implementation - `getCorrectString()` only removes null characters (`\u0000`):

```java
protected String getCorrectString(String value) {
    if (value != null) {
        return value.replace("\u0000", "");
    }
    return null;
}
```

It does not escape backslashes or any other characters, so LaTeX like `\frac{a}{b}` will pass through unchanged.

### 3. Inconsistent `isSupportedContent` Implementation

> Verify that `HtmlGenerator` has a similar content filtering mechanism

`HtmlGenerator` uses a different pattern - it doesn't have `isSupportedContent()`. Instead, unsupported content is handled by the `else` block in `write()` which simply returns without writing anything (line 143-145). `SemanticFormula` is already added to the `write()` method's dispatch chain.

---

All three issues don't require code changes. Let me know if you have other concerns!

Issue #198, opened 2026-02-19
Hybrid mode runs on CPU instead of GPU and fails with Invalid code point during preprocess stage


jsmount:
Hybrid mode runs on CPU instead of GPU and fails with Invalid code point

Hi,

I am trying to use Hybrid mode with opendataloader-pdf in Docker with
GPU support, but I am encountering two issues:

1.  The service runs on CPU instead of GPU
2.  The pipeline fails with Invalid code point during preprocessing

I would like to confirm whether my Docker setup is correct and ask for
more detailed guidance on properly configuring Hybrid mode.

  ------------
  Dockerfile
  ------------

FROM pytorch/pytorch:2.1.0-cuda12.1-cudnn8-runtime

RUN apt-get update && apt-get install -y curl libxcb1 libgl1
libglib2.0-0 && pip install “opendataloader-pdf[hybrid]” easyocr &&
apt-get clean

CMD [“opendataloader-pdf-hybrid”, “–port”, “5002”, “–host”, “0.0.0.0”,
“–force-ocr”]

  --------------------
  docker-compose.yml
  --------------------

services: hybrid: build: context: . dockerfile: Dockerfile.hybrid
ports: - “5002:5002” command: [“opendataloader-pdf-hybrid”, “–port”,
“5002”, “–host”, “0.0.0.0”, “–force-ocr”] environment: -
NVIDIA_VISIBLE_DEVICES=2 - CUDA_VISIBLE_DEVICES=2 deploy: resources:
reservations: devices: - driver: nvidia device_ids: [‘2’] capabilities:
[gpu] healthcheck: test: [“CMD”, “curl”, “-f”,
“http://localhost:5002/health”] interval: 10s timeout: 5s retries: 10
start_period: 60s

cli: image: ghcr.io/opendataloader-project/opendataloader-pdf-cli:1.10.1
depends_on: hybrid: condition: service_healthy volumes: -
./input:/input - ./output:/output command: > –hybrid docling-fast
–hybrid-url http://hybrid:5002 –hybrid-mode full –hybrid-timeout 600000
–replace-invalid-chars ? -f markdown -o /output /input/test_100.pdf

  -----------------------
  Issue 1: GPU Not Used
  -----------------------

Logs show:

UserWarning: ‘pin_memory’ argument is set as true but no accelerator is
found INFO - Accelerator device: ‘cpu’

Even though I set:

[...]

PR #216, opened 2026-02-23
fix: handle Docling PARTIAL_SUCCESS and fallback failed pages to Java

### LINKED ISSUE #215: fix: handle Docling PARTIAL_SUCCESS and report page-level errors in hybrid server
Derived from #198 (Issue 2: Invalid Code Point Error).

## Problem

When Docling encounters malformed Unicode in a PDF's font encoding during preprocessing, it raises an `Invalid code point` error for affected pages:

```
ERROR - Stage preprocess failed for run 1, pages [7]: Invalid code point
```

Docling handles this internally by skipping the failed page and returning `ConversionStatus.PARTIAL_SUCCESS`. However, the hybrid server currently:

1. Does not check `result.status` — always reports `"status": "success"` even when pages failed
2. Does not include `result.errors` — the Java client has no way to know which pages failed
3. The Java client cannot fall back to Java processing for the failed pages

## Expected Behavior

- The hybrid server should include Docling's conversion status and page-level errors in the response
- When status is `PARTIAL_SUCCESS`, the Java client should be able to identify failed pages and optionally reprocess them via the Java path
- The error should not cause the entire conversion to fail with a 500 error

## Affected Code

- `python/opendataloader-pdf/src/opendataloader_pdf/hybrid_server.py` — `convert_file()` endpoint
- `java/.../processors/HybridDocumentProcessor.java` — `processBackendPath()`

## Reproduction

Requires a PDF with malformed font encoding that triggers Docling's `Invalid code point` error during preprocessing (e.g., the reporter's `test_100.pdf` from #198).

hnc-leebd:
## Summary
Fixes #215 (derived from #198 Issue 2: Invalid Code Point Error)

- Python hybrid server now reports actual Docling conversion status (`partial_success`) instead of always `success`
- Java client parses `failed_pages` array from backend response
- Java processor automatically falls back to Java processing for pages that failed in the Docling backend

## Problem
When Docling encounters malformed Unicode in a PDF's font encoding during preprocessing, it raises `Invalid code point` and skips the affected pages, returning `PARTIAL_SUCCESS`. The hybrid server previously ignored this status, always reporting `"success"` — causing the Java client to silently drop failed pages with no content.

## Changes
- `hybrid_server.py` — New `build_conversion_response()` function that detects failed pages by comparing expected vs present page numbers; reports `status`, `errors`, and `failed_pages` in the response
- `HybridClient.java` — Added `failedPages` field to `HybridResponse` with backward-compatible constructors
- `DoclingFastServerClient.java` — Parses `partial_success` status and `failed_pages` array; logs warnings
- `HybridDocumentProcessor.java` — Converts failed page numbers from 1-indexed to 0-indexed, then falls back to Java processing path for those pages

## How to reproduce
```bash
# Using test PDF "1218_간질환정보집_최종본.pdf" (pages 1-10, page 7 fails)
opendataloader-pdf-hybrid  # start server
opendataloader-pdf --hybrid docling-fast --hybrid-mode full input.pdf
# Before fix: page 7 silently dropped
# After fix: page 7 falls back to Java processing
```

## How to test
```bash
# Java tests
cd java && mvn test -Dtest=DoclingFastServerClientTest

# Python tests
cd python/opendataloader-pdf && python -m pytest tests/test_hybrid_server_partial_success.py -v

# Full suite
./scripts/test-java.sh
cd python/opendataloader-pdf && python -m pytest tests/ -v
```

## Breaking changes
None. The response format is additive (new `errors`, `failed_pages` fields). Older Java clients that don't parse these fields continue to work. The `HybridResponse` class maintains backward-compatible constructors.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/hybrid/DoclingFastServerClient.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/hybrid/DoclingFastServerClient.java
@@ -18,8 +18,11 @@ import okhttp3.Response;
 import okhttp3.ResponseBody;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
@@ -155,10 +158,18 @@ public class DoclingFastServerClient implements HybridClient {
 
         // Check for API error status
         JsonNode statusNode = root.get("status");
-        if (statusNode != null && "failure".equals(statusNode.asText())) {
+        String status = statusNode != null ? statusNode.asText() : "";
+        if ("failure".equals(status)) {
             JsonNode errorsNode = root.get("errors");
             String errorMessage = errorsNode != null ? errorsNode.toString() : "Unknown error";
             throw new IOException("Docling Fast Server processing failed: " + errorMessage);
+        }
+
+        // Log partial_success status
+        if ("partial_success".equals(status)) {
+            JsonNode errorsNode = root.get("errors");
+            LOGGER.log(Level.WARNING, "Backend returned partial_success: {0}",
+                errorsNode != null ? errorsNode.toString() : "no error details");
         }
 
         // Extract document content
@@ -172,7 +183,10 @@ public class DoclingFastServerClient implements HybridClient {
         // Extract per-page content from json_content if available
         Map<Integer, JsonNode> pageContents = extractPageContents(jsonContent);
 
-        return new HybridResponse(null, null, jsonContent, pageContents);
+        // Extract failed pages (1-indexed) from partial_success responses
+        List<Integer> failedPages = extractFailedPages(root);
+
+        return new HybridResponse(null, null, jsonContent, pageContents, failedPages);
     }
 
     /**
@@ -205,6 +219,28 @@ public class DoclingFastServerClient implements HybridClient {
         }
 
         return pageContents;
+    }
+
+    /**
+     * Extracts the list of failed page numbers from the response.
+     *
+     * <p>When the backend returns partial_success, the failed_pages array contains
+     * 1-indexed page numbers that failed during processing (e.g., due to Invalid code point
+     * errors in PDF font encoding).
+     */
+    private List<Integer> extractFailedPages(JsonNode root) {
+        JsonNode failedPagesNode = root.get("failed_pages");
+        if (failedPagesNode == null || !failedPagesNode.isArray() || failedPagesNode.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        List<Integer> failedPages = new ArrayList<>();
+        for (JsonNode pageNode : failedPagesNode) {
+            if (pageNode.isNumber() && pageNode.canConvertToInt()) {
+                failedPages.add(pageNode.asInt());
+            }
+        }
+        return failedPages;
     }
 
     /**
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/hybrid/HybridClient.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/hybrid/HybridClient.java
@@ -10,9 +10,11 @@ package org.opendataloader.pdf.hybrid;
 import com.fasterxml.jackson.databind.JsonNode;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -176,6 +178,7 @@ public interface HybridClient {
         private final String html;
         private final JsonNode json;
         private final Map<Integer, JsonNode> pageContents;
+        private final List<Integer> failedPages;
 
         /**
          * Creates a new HybridResponse.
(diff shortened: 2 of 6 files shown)


hnc-leebd:
## Review Response

Evaluated each item against the codebase. Accepted 5, declined 6 with reasoning below.

### Accepted & Fixed (commit 4580e55)

| # | Item | Fix |
|---|------|-----|
| 1 | Mutable list from `getFailedPages()` | Wrapped in `Collections.unmodifiableList(new ArrayList<>(...))` in HybridResponse constructor |
| 2 | `isInt()` too narrow for Jackson | Changed to `isNumber() && canConvertToInt()` in `extractFailedPages()` |
| 3 | Boundary detection blind spot | Added `logger.warning()` when min/max heuristic fallback is used |
| 7 | Missing all-pages-fail test | Added 2 Python tests and 1 Java test covering empty pages dict |
| 10 | Duplicate log messages | Removed WARNING from `processBackendPath`, consolidated into single WARNING in `processDocument` |

### Declined with Reasoning

**4. Output parameter pattern → record**: `processBackendPath` is a private method with clear Javadoc on the output parameter. A record adds a new type for a single call site. The current pattern is straightforward and well-documented.

**5. Python 3.10+ union syntax**: `pyproject.toml` declares `requires-python = ">=3.10"`. The `tuple[int, int] | None` syntax is valid for the project's minimum version.

**6. String-based status constants**: These strings come from `result.status.value` (Docling's enum). Module-level constants for two usages in a single file adds indirection. If Docling changes its enum values, we'd need to update the constants the same way we'd update inline strings.

**8. Fragile exception assertion**: The `"processing failed"` substring is part of our own error message in `DoclingFastServerClient.parseResponse()`, not an external dependency. We control this string.

**9. Missing HybridDocumentProcessor integration test**: Mocking the full processing chain (`Config`, `IDocument`, content filtering, page heights) has high setup cost. The 1-indexed to 0-indexed conversion is a single arithmetic operation (`- 1`) with a bounds check. The boundary is already tested: client tests verify failed page parsing, Java path tests verify processing.

**11. `errors` field in success responses**: Additive field (`"errors": []`). The Java client's `extractFailedPages()` handles missing fields gracefully. No downstream consumer checks for absence of `errors` to infer status — they check the `"status"` field.

hnc-leebd:
## Review Response (Round 2)

Evaluated 5 items. Accepted 3, declined 2.

### Accepted & Fixed (commit a2e796a)

| # | Item | Fix |
|---|------|-----|
| 2 | Log 0-indexed page numbers confuse users | Log message now shows 1-indexed pages via `stream().map(p -> p + 1)` |
| 3 | `int(k)` can raise ValueError | Wrapped in try/except with warning log for non-integer keys |
| 5 | Stringly-typed status comparison | `convert_file` endpoint now uses `result.status == ConversionStatus.PARTIAL_SUCCESS` directly. `build_conversion_response` retains string comparison intentionally — it's a standalone function tested without docling dependency. |

### Declined with Reasoning

**1. Output parameter risks uninitialized `backendResults`**: Traced the control flow at lines 150–162. If `processBackendPath` throws after populating `backendFailedPages`, the catch block either (a) assigns `backendResults = processJavaPath(...)` when fallback is enabled, or (b) re-throws when fallback is disabled. There is no code path where `backendResults` is uninitialized at line 165. The record refactoring was also raised in round 1 and declined for the same reason — private method, single call site, clear Javadoc.

**4. `errors` attribute assumption**: Verified against Docling source — `ErrorItem` is a Pydantic `BaseModel` with `error_message: str` as a required field. The `getattr` fallback to `str(e)` is already defensive, and `str()` on a Pydantic model produces a readable representation, not `<object at 0x...>`.

hnc-leebd:
## Review Response (Round 3)

Evaluated 6 items. Accepted 1, declined 5.

### Accepted

| # | Item | Action |
|---|------|--------|
| 1 | Per-page fallback ignores `fallbackToJava` config | Will fix — wrap the per-page fallback block in the same `isFallbackToJava()` guard. When disabled, log a warning with the failed page numbers but skip the Java fallback, preserving whatever content the backend returned for the successful pages. |

### Declined with Reasoning

**2. Output parameter → record**: Declined in rounds 1 and 2 for the same reason. `processBackendPath` is a private method with a single call site. The mutable `Set` parameter is documented in Javadoc. A record wrapper adds a type for one use.

**3. ConversionStatus imported inside function body**: All `docling` imports in `hybrid_server.py` are inside function bodies — this is a deliberate pattern because `docling` is an optional dependency (`pip install opendataloader-pdf[hybrid]`). See lines 178, 210–218 for the same pattern. Moving it to top-level would cause `ImportError` at module load when the hybrid extra is not installed.

**4. Inconsistent status comparison (enum vs string)**: By design. `convert_file` uses enum comparison (`ConversionStatus.PARTIAL_SUCCESS`) because it has access to the docling runtime. `build_conversion_response` takes `status_value: str` intentionally — it's a standalone, independently-testable function with no docling dependency. The string parameter is part of its API contract.

**5. Python 3.10+ type hints**: Declined in round 1. `pyproject.toml` declares `requires-python = ">=3.10"`. The `tuple[int, int] | None` syntax is valid.

**6. Missing HybridDocumentProcessor integration test**: Declined in round 2. The 1-indexed → 0-indexed conversion is a single arithmetic operation with a bounds check. Client tests verify failed page parsing, Java path tests verify processing. The join point is a three-line `if` block.

hnc-leebd:
## Review Response (Round 4)

Evaluated 6 items. Accepted 2, declined 4.

### Accepted & Fixed (commit b91b909)

| # | Item | Fix |
|---|------|-----|
| 5 | Missing test for malformed `failed_pages` values | Added Java test with `[3, "bad", null, 5]` — verifies only valid integers are extracted |
| 5 | Missing tests for `failure` status and missing `pages` key | Added 2 Python tests: `failure` status produces empty `failed_pages`; missing `pages` key with `partial_success` reports all requested pages as failed |
| 6 | `total_pages=0` is falsy | Changed `elif total_pages:` to `elif total_pages is not None:` |

### Declined with Reasoning

**1. Status string coupling between `convert_file` and `build_conversion_response`**: Declined in round 3 (item 4). By design — `build_conversion_response` is a standalone function with no docling dependency, accepting `status_value: str` as its API contract. The enum comparison lives in `convert_file` which has the docling runtime. These serve different architectural concerns.

**2. Import inside request handler**: Declined in round 3 (item 3). All docling imports in `hybrid_server.py` are inside function bodies because docling is an optional dependency (`[hybrid]` extra). Lines 178, 210–218 follow the same pattern. Top-level import would cause `ImportError` when the extra is not installed.

**3. `page_count` attribute assumption**: `getattr(result.input, "page_count", None)` already returns `None` defensively when the attribute is absent. The `None` case is handled by the boundary-detection fallback path (with warning log, added in round 1). No silent failure — the code degrades gracefully.

**4. Output parameter pattern**: Declined in rounds 1, 2, and 3. Private method, single call site, documented in Javadoc.

PR #255, opened 2026-03-05
fix(markdown): preserve split for fused numbered heading subtitles

### LINKED ISSUE #254: Preserve heading line break when numbered heading is fused with hyphenated subtitle
## Problem
For some academic PDFs, a heading line that is visually split into two lines in source appears fused in markdown output.

Example fused output:
##### 1.1. ContributionsPost-Training: Large-Scale Reinforcement Learning on the Base Model

Expected shape:
- ##### 1.1. Contributions
- Post-Training: Large-Scale Reinforcement Learning on the Base Model

## Repro
Observed with DeepSeek-R1 PDF using default markdown output (also with use_struct_tree=true and keep_line_breaks=true).

## Proposal
Add a targeted markdown heading normalization for numbered headings where a short heading label is fused with a likely subtitle token like Post-...:.

I prepared a PR with a small, conservative fix plus tests.

StevenVincentOne:
## Summary
- add a targeted split for fused numbered heading + hyphenated subtitle patterns in markdown generation
- handle cases like 1.1. ContributionsPost-Training: ... by emitting:
  - heading: 1.1. Contributions
  - following line: Post-Training: ...
- add unit tests for split and non-split cases

## Notes
- verified on DeepSeek-R1 sample where source layout had a visible line break but markdown heading was fused
- no behavior change for normal numbered headings (e.g. 2.2.1 Reinforcement Learning Algorithm)

Closes #254

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/markdown/MarkdownGenerator.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/markdown/MarkdownGenerator.java
@@ -34,6 +34,8 @@ import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class MarkdownGenerator implements Closeable {
 
@@ -46,6 +48,9 @@ public class MarkdownGenerator implements Closeable {
     protected boolean embedImages = false;
     protected String imageFormat = Config.IMAGE_FORMAT_PNG;
     protected boolean includeHeaderFooter = false;
+    private static final Pattern NUMBERED_HEADING_RE = Pattern.compile("^(\\d+(?:\\.\\d+)+\\.?)\\s+(.+)$");
+    private static final Pattern FUSED_SUBTITLE_RE =
+        Pattern.compile("^(.+?)\\s+([A-Z][A-Za-z0-9]+(?:-[A-Za-z0-9]+)+:\\s+.+)$");
 
     MarkdownGenerator(File inputPdf, Config config) throws IOException {
         String cutPdfFileName = inputPdf.getName();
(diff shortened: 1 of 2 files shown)


bundolee (commented):
Thank you for the contribution and for filing #254 — the fused heading problem is real and worth fixing.

However, I have some concerns about the approach that I'd like to discuss before we proceed.

### Wrong layer for the fix

The root cause is that line-break information is lost during PDF text extraction / structure tree parsing, before `SemanticHeading` is even constructed. `MarkdownGenerator` is a rendering layer — it should translate semantic nodes into Markdown syntax, not re-analyze and split text content.

Fixing this here sets a precedent where the renderer compensates for upstream parsing issues with pattern-matching heuristics, which will be hard to maintain over time.

### Narrow heuristic

The regex and the hardcoded word list (`Post`, `Pre`, `Self`, `Co`) effectively target a single document (DeepSeek-R1). Other fused headings that don't match this pattern — different casing, no hyphen, no colon — would still be broken. A fix at the parsing layer would handle all of these generically.

### Bypasses existing code path

The original `writeHeading()` delegates to `writeSemanticTextNode()`, which handles line-break normalization for `keepLineBreaks` mode and table contexts. The PR replaces this with direct `getCorrectMarkdownString()` calls and partially duplicates the line-break logic at the top of the method. This divergence could introduce subtle regressions that the two unit tests wouldn't catch.

### Suggestion

I'd like to address this at the text extraction / chunk merging stage, where line-break information from the source PDF is still available. That way the fix is generic rather than pattern-specific.

I'll keep #254 open and use it to track the proper fix. Thanks again for identifying the issue and putting together a clean PR — it helped pinpoint exactly where the problem manifests.

StevenVincentOne:
Thanks for the detailed review. I agree with the main point: this fix is in the wrong layer.

The current PR solves a real symptom, but it does it in `MarkdownGenerator`, and that is not the right long-term place to repair fused heading text. I do not want to normalize parser/extraction issues in the renderer.

I’m going to leave the underlying problem tracked in #254. If I come back to it, I’ll do it at the text extraction / chunk merging stage instead of in markdown rendering.

Thanks again for the review and for calling out the architectural boundary clearly.

StevenVincentOne:
Closing this PR as discussed. The underlying issue remains valid, but the fix belongs earlier in the pipeline at text extraction / chunk merging rather than in `MarkdownGenerator`.

I’m leaving the problem tracked in #254 and would revisit it there with a parsing-layer approach if I come back to it.

Thanks again for the review.

PR #266, opened 2026-03-06
Fix propagate StaticContainers ThreadLocal context for hybrid page workers

### LINKED ISSUE #264: bug: verapdf StaticContainers uses ThreadLocal — blocks per-page parallel processing
## Summary

All fields in verapdf's `StaticContainers` class are declared as `ThreadLocal`. When per-page processing is executed on `ForkJoinPool` worker threads, these values are `null`, causing `NullPointerException`.

## Root Cause

`StaticContainers` field structure (verapdf `wcag-algorithms` library):

```java
private static final ThreadLocal<Boolean> isIgnoreCharactersWithoutUnicode;
private static final ThreadLocal<Boolean> isDataLoader;
private static final ThreadLocal<IDocument> document;
private static final ThreadLocal<TableBordersCollection> tableBordersCollection;
// ... all fields are ThreadLocal
```

`DocumentProcessor.preprocessing()` initializes these on the **main thread only**:

```java
// DocumentProcessor.java:287-288
StaticContainers.setIsDataLoader(true);
StaticContainers.setIsIgnoreCharactersWithoutUnicode(false);
```

When `TextLineProcessor.processTextLines()` or similar methods are called from a different thread, the `ThreadLocal` values are `null` → NPE.

## Stack Trace

```
java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()"
  because the return value of "StaticContainers.getIsIgnoreCharactersWithoutUnicode()" is null
    at TextChunk.isEmpty(TextChunk.java:291)
    at TextLineProcessor.processTextLines(TextLineProcessor.java:38)
```

## Benchmark Impact (684 PDFs)

| Metric | Value |
|--------|-------|
| Documents failing with NPE | ~57% (documents with 8+ pages) |
| Quality of successful documents | Identical to baseline (NID/TEDS/MHS diff = 0) |
| Speed improvement (successful only) | -3~5% |

## Affected Code

Code depending on `StaticContainers`:

- `DocumentProcessor.java` — calls `TextLineProcessor`, `ParagraphProcessor`
- `HybridDocumentProcessor.java` — same pattern
- `TextChunk.isEmpty()` — directly references `StaticContainers.getIsIgnoreCharactersWithoutUnicode()`
- `TableBorderProcessor` — references `StaticContainers.getTableBordersCollection()`
- `MarkdownGenerator` — references `StaticContainers.getDocument()`, `isKeepLineBreaks()`

## Proposed Solutions

1. **ThreadLocal propagation**: Capture ThreadLocal values on main thread, re-set them on worker threads
   ```java
   boolean ignoreChars = StaticContainers.getIsIgnoreCharactersWithoutUnicode();
   boolean isDataLoader = StaticContainers.getIsDataLoader();
   // On worker thread entry:
   StaticContainers.setIsIgnoreCharactersWithoutUnicode(ignoreChars);
   StaticContainers.setIsDataLoader(isDataLoader);
   ```
2. **Custom ForkJoinPool**: Include StaticContainers initialization in ThreadFactory
3. **Upstream verapdf fix**: Request change from ThreadLocal to plain static fields or InheritableThreadLocal

SergioChan:
**Issue resolved by this Pull Request:**
Resolves #264

**Summary of changes:**
- Add `StaticContainersThreadContext` utility to capture/apply verapdf `StaticContainers` ThreadLocal-backed values on worker threads.
- Update `HybridDocumentProcessor.processJavaPath` to process pages via `parallelStream()` and re-apply captured context per worker before table/text processing.
- Add regression test `StaticContainersThreadContextTest` covering worker-thread application before `TextLineProcessor.processTextLines`.

**Validation:**
- Attempted: `mvn -pl opendataloader-pdf-core -Dtest=StaticContainersThreadContextTest,TextLineProcessorTest test`
- Result: build cannot run in this environment because local JDK does not support project-required release target (`error: release version 11 not supported`).

**Checklist:**

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/HybridDocumentProcessor.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/HybridDocumentProcessor.java
@@ -37,6 +37,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -265,25 +266,42 @@ public class HybridDocumentProcessor {
             new ClusterTableProcessor().processTables(workingContents);
         }
 
-        // Process each page through the standard Java pipeline.
-        // StaticContainers uses ThreadLocal fields, so worker threads must inherit context explicitly.
+        // Stage 1 (sequential): table-border processing mutates shared table-border collections.
+        // Keep it on the main thread to avoid concurrent mutation of shared verapdf state.
+        for (int pageNumber : pageNumbers) {
+            List<IObject> pageContents = workingContents.get(pageNumber);
+            pageContents = TableBorderProcessor.processTableBorders(pageContents, pageNumber);
+            pageContents = pageContents.stream()
+                .filter(x -> !(x instanceof LineChunk))
+                .collect(Collectors.toList());
+            workingContents.set(pageNumber, pageContents);
+        }
+
+        // Stage 2 (parallel): text-line processing with ThreadLocal flag propagation.
+        // Results are collected in a concurrent map instead of mutating ArrayList from worker threads.
         StaticContainersThreadContext.Snapshot threadContext = StaticContainersThreadContext.capture();
+        Map<Integer, List<IObject>> textLineResults = new ConcurrentHashMap<>();
         pageNumbers.parallelStream().forEach(pageNumber -> {
             try {
                 StaticContainersThreadContext.apply(threadContext);
                 List<IObject> pageContents = workingContents.get(pageNumber);
-                pageContents = TableBorderProcessor.processTableBorders(pageContents, pageNumber);
-                pageContents = pageContents.stream()
-                    .filter(x -> !(x instanceof LineChunk))
-                    .collect(Collectors.toList());
                 pageContents = TextLineProcessor.processTextLines(pageContents);
-                pageContents = SpecialTableProcessor.detectSpecialTables(pageContents);
-                workingContents.set(pageNumber, pageContents);
+                textLineResults.put(pageNumber, pageContents);
             } catch (Exception e) {
                 LOGGER.log(Level.WARNING, "Error processing page {0}: {1}",
                     new Object[]{pageNumber, e.getMessage()});
+                textLineResults.put(pageNumber, workingContents.get(pageNumber));
+            } finally {
+                StaticContainersThreadContext.clear();
             }
         });
+
+        // Stage 3 (sequential): special-table detection touches table-border structures.
+        for (int pageNumber : pageNumbers) {
+            List<IObject> pageContents = textLineResults.getOrDefault(pageNumber, workingContents.get(pageNumber));
+            pageContents = SpecialTableProcessor.detectSpecialTables(pageContents);
+            workingContents.set(pageNumber, pageContents);
+        }
 
         // Apply cross-page processing for Java pages only
         applyJavaPagePostProcessing(workingContents, pageNumbers);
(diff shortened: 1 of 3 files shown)


SergioChan:
Thanks for the thorough review �� I pushed a follow-up in 87f830b addressing the concurrency concerns.

### What changed
1. **Removed shared mutable table-border mutations from parallel workers**
   - `TableBorderProcessor.processTableBorders(...)` now runs in a **sequential stage** before parallel work.
   - `SpecialTableProcessor.detectSpecialTables(...)` now runs in a **sequential stage** after parallel text-line processing.

2. **Avoided concurrent writes to `workingContents`**
   - Parallel stage now writes page outputs to a `ConcurrentHashMap<Integer, List<IObject>>`.
   - Main thread merges those results back into `workingContents` afterward.

3. **Narrowed ThreadLocal propagation scope**
   - `StaticContainersThreadContext` now propagates only immutable processing flags needed for worker text-line processing.
   - Shared objects (`document`, `tableBordersCollection`) are no longer propagated into workers.

4. **Added worker ThreadLocal cleanup**
   - Added `StaticContainersThreadContext.clear()` and call it in a `finally` block inside the parallel worker.

5. **Null/stale state handling**
   - `capture()` now fails fast via `Objects.requireNonNull(...)` if required flags were not initialized, instead of silently skipping state application.

6. **Updated test coverage**
   - Extended `StaticContainersThreadContextTest` to:
     - include `keepLineBreaks` in captured/applied context
     - verify cleanup behavior on a reused worker thread

### Validation
I could not run the Maven test target here because `mvn` is not available in this environment, but the changes are scoped and aligned with the concerns raised.

bundolee:
Hi @SergioChan, thanks for the effort on this PR and especially the thorough follow-up in 87f830b. Your work — and the review discussion that followed — led us to dig much deeper into veraPDF's threading model than we had before, and what we found changes the picture significantly.

**What your PR helped us discover:**

Through reviewing your Snapshot capture/apply pattern, we asked ourselves: "are these 3 boolean flags everything a worker thread needs?" That question led to a full audit of veraPDF's `StaticContainers` ecosystem, and the answer turned out to be no — not by a wide margin.

### The fundamental problem: veraPDF is single-threaded by design

veraPDF's container classes hold **47 ThreadLocal fields across 6 classes**, of which **34 are mutable objects** (PDDocument, font caches, TableBordersCollection, AccumulatedNodeMapper, etc.). There is zero synchronization anywhere in the veraPDF codebase — no `synchronized`, no `volatile`, no `ConcurrentHashMap`.

This PR propagates 3 boolean flags. The other 44 fields remain null or stale on worker threads. Even if we propagated all 47, we'd be sharing **references** to the same mutable objects across threads without synchronization, which is fundamentally unsafe.

### Why testing can't prove safety

The unit test passes because it exercises a narrow path that doesn't trigger deeper veraPDF internals. Concurrency bugs here are non-deterministic — they depend on CPU core count, ForkJoinPool scheduling, and PDF content. A passing CI test doesn't rule out silent data corruption in production.

### Cost-benefit

The only parallelized stage is `TextLineProcessor.processTextLines()`. TableBorder and SpecialTable processing both had to stay sequential. In hybrid mode, the bottleneck is typically the Python backend call, not the Java page loop. Without profiling data showing this is a meaningful bottleneck, the complexity (3-stage pipeline, ThreadLocal propagation, ConcurrentHashMap, cleanup) isn't justified.

### Path forward

I'm closing this PR, but the analysis it prompted is genuinely valuable. If page-level parallelism becomes a demonstrated need, safer approaches would be:

- **Document-level parallelism** — fully isolated StaticContainers state per thread
- **Process-level isolation** — separate JVM processes
- **Upstream contribution** — work with veraPDF to add thread-safety

Thank you for the contribution. This is one of those cases where the investigation was more valuable than the outcome — we now have a much clearer understanding of what's safe and what isn't.

bundolee:
Closing — see comment above for detailed rationale.

PR #274, opened 2026-03-11
Skip malformed cluster tables to avoid full job crash

### LINKED ISSUE #238: [Bug] IndexOutOfBoundsException in ClusterTableProcessor crashes entire PDF processing job
### Bug

When converting a large PDF using `table_method="cluster"`, the processing crashes
with a `java.lang.IndexOutOfBoundsException: Index 2 out of bounds for length 2`
thrown inside `Table.createTableBorderFromTable()` at line 350. This terminates
the entire job — no output is produced for any page.

The error originates in `ClusterTableProcessor.processClusterDetectionTables()`
when the cluster algorithm encounters a specific table structure it cannot handle,
attempting to access index 2 of a list with only 2 elements (valid indices: 0–1).

The only current workaround is switching to `table_method="default"`, which loses
cluster-based detection entirely.

...

### Steps to reproduce

1. Install the library:
pip install opendataloader-pdf==1.11.2

2. Run the following Python code against a large PDF (1000+ pages):
```python
import opendataloader_pdf

opendataloader_pdf.convert(
    input_path="path/to/large.pdf",
    output_dir="path/to/output",
    image_dir="path/to/output/images",
    format="json,text,html,pdf,markdown,markdown-with-html,markdown-with-images",
    table_method="cluster"
)

3. Observe the following exception in the output:

SEVERE: Exception during processing file path/to/large.pdf:
Index 2 out of bounds for length 2
java.lang.IndexOutOfBoundsException: Index 2 out of bounds for length 2
    at java.base/java.util.Objects.checkIndex(Objects.java:365)
    at java.base/java.util.ArrayList.get(ArrayList.java:428)
    at org.verapdf.wcag.algorithms.entities.tables.Table.createTableBorderFromTable(Table.java:350)
    at org.opendataloader.pdf.processors.ClusterTableProcessor.processClusterDetectionTables(ClusterTableProcessor.java:64)
    at org.opendataloader.pdf.processors.ClusterTableProcessor.getTables(ClusterTableProcessor.java:32)
    at org.opendataloader.pdf.processors.AbstractTableProcessor.processTables(AbstractTableProcessor.java:49)
    at org.opendataloader.pdf.processors.AbstractTableProcessor.processTables(AbstractTableProcessor.java:38)
    at org.opendataloader.pdf.processors.DocumentProcessor.processDocument(DocumentProcessor.java:140)
    at org.opendataloader.pdf.processors.DocumentProcessor.processFile(DocumentProcessor.java:76)
    at org.opendataloader.pdf.cli.CLIMain.processFile(CLIMain.java:113)

**Expected behavior**
Either:
- The processor skips the problematic table/page and continues with the rest of the document, OR
- A meaningful error message is raised indicating which page/table caused the issue

**Actual behavior**
The entire job crashes and no output files are produced.

**Possible root cause**
Table.createTableBorderFromTable() at line 350 assumes the list has at least 3
elements before calling .get(2), but does not guard against lists of length < 3.
A bounds check or a try/catch around this call would prevent the crash.

**Workaround used**
Switched to table_method="default" until this is fixed.
...

### Version

- opendataloader-pdf -> 1.11.2
- python -> 3.11
- OS: wINDOWS 10
- PDF: large document (1272+ pages), contains pages with background elements
  (WARNING: Detected background on page 1272 appears before the crash)

...

### Java version

java 25.0.2 2026-01-20 LTS
Java(TM) SE Runtime Environment (build 25.0.2+10-LTS-69)
Java HotSpot(TM) 64-Bit Server VM (build 25.0.2+10-LTS-69, mixed mode, sharing)

...

<!-- ⚠️ ATTENTION: When sharing screenshots, attachments, or other data make sure not to include any sensitive information. -->

SergioChan:
## Summary
- prevent a single malformed cluster-detected table from aborting the whole page/document flow
- wrap `table.createTableBorderFromTable()` in `ClusterTableProcessor` with an `IndexOutOfBoundsException` guard
- log a warning and continue processing remaining detected tables
- add `ClusterTableProcessorTest` to verify malformed tables are skipped while valid tables are still emitted

## Why
Issue #238 reports `IndexOutOfBoundsException` from `Table.createTableBorderFromTable()` crashing conversion jobs when `table_method="cluster"` is used on large PDFs.

This change keeps processing resilient by skipping only the bad table instance instead of failing the entire run.

Closes #238

## Validation
- Added focused unit test: `ClusterTableProcessorTest#testCollectTableBordersSkipsMalformedTableAndContinues`
- Local Maven execution was not possible in this environment (`mvn` is unavailable).

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/ClusterTableProcessor.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/ClusterTableProcessor.java
@@ -26,12 +26,16 @@ import org.verapdf.wcag.algorithms.semanticalgorithms.utils.TextChunkUtils;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Table processor that uses clustering algorithms to detect tables.
  * Identifies tables by analyzing spatial relationships between text chunks.
  */
 public class ClusterTableProcessor extends AbstractTableProcessor {
+
+    private static final Logger LOGGER = Logger.getLogger(ClusterTableProcessor.class.getCanonicalName());
 
     @Override
     protected List<List<TableBorder>> getTables(List<List<IObject>> contents, List<Integer> pageNumbers) {
@@ -67,11 +71,19 @@ public class ClusterTableProcessor extends AbstractTableProcessor {
             }
         }
         clusterTableConsumer.processEnd();
+        return collectTableBorders(clusterTableConsumer.getTables());
+    }
+
+    static List<TableBorder> collectTableBorders(List<Table> tables) {
         List<TableBorder> result = new ArrayList<>();
-        for (Table table : clusterTableConsumer.getTables()) {
-            TableBorder tableBorder = table.createTableBorderFromTable();
-            if (tableBorder != null) {
-                result.add(tableBorder);
+        for (Table table : tables) {
+            try {
+                TableBorder tableBorder = table.createTableBorderFromTable();
+                if (tableBorder != null) {
+                    result.add(tableBorder);
+                }
+            } catch (IndexOutOfBoundsException exception) {
+                LOGGER.log(Level.WARNING, () -> "Skipping malformed cluster-detected table: " + exception.getMessage());
             }
         }
         return result;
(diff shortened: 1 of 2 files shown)


MaximPlusov:
Thank you for the contribution.

But the original issue https://github.com/opendataloader-project/opendataloader-pdf/issues/238 fixed within the veraPDF library.

SergioChan:
Thanks for the clarification and for pointing to the upstream veraPDF fix. That makes sense — I’ll close this PR so we don’t duplicate work here.

If a related gap reappears in this repo, I’m happy to reopen with a targeted follow-up.

PR #278, opened 2026-03-11
fix: change hybrid_fallback default to false for fail-fast behavior

### LINKED ISSUE #277: fix: hybrid mode silently falls back to Java when backend server is unavailable
## Problem

When `--hybrid docling-fast` is specified but the backend server is not running, the system **silently falls back to Java-only processing** instead of failing with an error. This makes it impossible for users and automated agents to detect that hybrid processing was not actually performed.

The output appears successful, but the quality is degraded compared to what hybrid mode would produce.

## Root Cause

`hybrid_fallback` defaults to `true` across all layers:

| Layer | Location | Default |
|-------|----------|---------|
| Java | `HybridConfig.java:43` — `fallbackToJava = true` | `true` |
| CLI | `CLIOptions.java:177` — `OptionDefinition` default | `true` |
| Python | `convert_generated.py:37` — `hybrid_fallback: bool = True` | `true` |

### Fallback flow

1. User passes `--hybrid docling-fast` (without explicit `--hybrid-fallback`)
2. Python wrapper adds `--hybrid-fallback` by default (`hybrid_fallback=True`)
3. Health check at `HybridDocumentProcessor.java:112` throws `IOException`
4. Exception caught at `CLIMain.java:122` — logged as SEVERE but process exits with code 0
5. **Additionally**, even when the health check passes but backend processing fails mid-flight (`HybridDocumentProcessor.java:167-174`), the fallback silently processes all backend pages through the Java path

In both cases, the caller receives no indication that hybrid processing was not performed.

## Expected Behavior

- **Default** (`--hybrid docling-fast` without `--hybrid-fallback`): Server unavailable → **fail with error**
- **Explicit opt-in** (`--hybrid docling-fast --hybrid-fallback`): Server unavailable → graceful degradation to Java processing (current behavior)

## Proposed Fix

Change `hybrid_fallback` default from `true` to `false` (fail-fast by default):

1. `HybridConfig.java:43` — `private boolean fallbackToJava = false;`
2. `CLIOptions.java:177` — change default to `false`
3. `CLIOptions.java:131` — update description: `"Default: false"`
4. `options.json` — update default to `false`
5. Run `npm run sync` to regenerate Python/Node.js bindings

## Impact

- Users who explicitly want fallback behavior must now pass `--hybrid-fallback`
- Users who expect hybrid quality will get a clear error when the server is down
- Automated pipelines can properly detect and handle hybrid server failures

bundolee:
## Summary
- Change `hybrid_fallback` default from `true` to `false` across all layers (Java, Python, Node.js, CLI)
- When `--hybrid` is requested but the server is unavailable, execution now **fails with an error** instead of silently falling back to Java-only processing
- Users who want graceful degradation can explicitly opt in with `--hybrid-fallback`

Fixes #277

## Problem
When `--hybrid docling-fast` was specified without an available backend server, the system silently fell back to Java-only processing and returned success. This made it impossible for users and automated agents to detect that hybrid processing was not actually performed, leading to undetected quality degradation.

## Changes
- `java/.../hybrid/HybridConfig.java` — `fallbackToJava` default: `true` → `false`
- `java/.../cli/CLIOptions.java` — option definition default and description updated
- `options.json` — source of truth default updated
- Auto-generated bindings (Python, Node.js, docs) regenerated via `npm run sync`
- Tests updated to verify new default behavior

## How to test
```bash
# Without --hybrid-fallback (default): should fail if server is down
opendataloader-pdf --hybrid docling-fast test.pdf
# Expected: error "Hybrid server is not available at http://localhost:5002"

# With explicit --hybrid-fallback: graceful degradation (previous behavior)
opendataloader-pdf --hybrid docling-fast --hybrid-fallback test.pdf
# Expected: processes with Java fallback, logs warning
```

## Breaking changes
Users relying on silent fallback behavior will now get errors when the hybrid server is unavailable. Add `--hybrid-fallback` to restore previous behavior.

### DIFF
--- a/content/docs/license.mdx
+++ b/content/docs/license.mdx
@@ -3,12 +3,12 @@ title: License
 description: License information for OpenDataLoader PDF
 ---
 
-OpenDataLoader PDF is released under the [Mozilla Public License 2.0](https://www.mozilla.org/MPL/2.0/). See the repository for additional details:
+OpenDataLoader PDF is released under the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). See the repository for additional details:
 
 - `LICENSE`
+- `NOTICE`
 - `THIRD_PARTY/THIRD_PARTY_LICENSES.md`
 - `THIRD_PARTY/THIRD_PARTY_NOTICES.md`
-- `THIRD_PARTY/licenses/`
 
 ## Summary
 


bundolee:
Thanks for the thorough review! All three points are reasonable — here's the plan:

**Issue 1 (license.mdx): Accepted** — Will remove the unrelated `content/docs/license.mdx` hunk from this PR and open a dedicated PR for it.

**Issue 2 (integration test for fail-fast error path): Accepted** — Will add a test in `HybridDocumentProcessorTest` that mocks an unavailable server and asserts an exception is thrown when `fallbackToJava=false`.

**Issue 3 (Python docstring capitalization): Accepted** — Will update the description in `options.json` to use `"Default: False"` (Python convention) and re-run `npm run sync`.

Will push the fixes shortly.

PR #280, opened 2026-03-12
fix: change hybrid_fallback default to false for fail-fast behavior

### LINKED ISSUE #277: fix: hybrid mode silently falls back to Java when backend server is unavailable

bundolee:
## Summary
- Change `hybrid_fallback` default from `true` to `false` across all layers (Java, Python, Node.js, CLI)
- When `--hybrid` is requested but the server is unavailable, execution now **fails with an error** instead of silently falling back to Java-only processing
- Users who want graceful degradation can explicitly opt in with `--hybrid-fallback`

Fixes #277

## Problem
When `--hybrid docling-fast` was specified without an available backend server, the system silently fell back to Java-only processing and returned success. This made it impossible for users and automated pipelines to detect that hybrid processing was not actually performed.

## Changes
- `HybridConfig.java` — `fallbackToJava` default: `true` → `false`
- `CLIOptions.java` — option description updated
- `options.json` — source of truth default updated
- Auto-generated bindings (Python, Node.js, docs) regenerated via `npm run sync`
- Tests added for `checkAvailability()` error paths (server unavailable, unhealthy server)
- Existing tests updated to verify new default behavior

## How to test
```bash
# Without --hybrid-fallback (default): should fail if server is down
opendataloader-pdf --hybrid docling-fast test.pdf
# Expected: error "Hybrid server is not available at http://localhost:5002"

# With explicit --hybrid-fallback: graceful degradation (previous behavior)
opendataloader-pdf --hybrid docling-fast --hybrid-fallback test.pdf
# Expected: processes with Java fallback, logs warning
```

## Breaking changes
Users relying on silent fallback behavior will now get errors when the hybrid server is unavailable. Add `--hybrid-fallback` to restore previous behavior.

### DIFF
--- a/content/docs/_generated/node-convert-options.mdx
+++ b/content/docs/_generated/node-convert-options.mdx
@@ -32,4 +32,4 @@ description: Options for the Node.js convert function
 | `hybridMode`            | `string`             | `"auto"`     | Hybrid triage mode. Values: auto (default, dynamic triage), full (skip triage, all pages to backend)                               |
 | `hybridUrl`             | `string`             | -            | Hybrid backend server URL (overrides default)                                                                                      |
 | `hybridTimeout`         | `string`             | `"30000"`    | Hybrid backend request timeout in milliseconds. Default: 30000                                                                     |
-| `hybridFallback`        | `boolean`            | `true`       | Fallback to Java processing on hybrid backend error. Default: true                                                                 |
+| `hybridFallback`        | `boolean`            | `false`      | Opt in to Java fallback on hybrid backend error (default: disabled)                                                                |
--- a/content/docs/_generated/python-convert-options.mdx
+++ b/content/docs/_generated/python-convert-options.mdx
@@ -33,4 +33,4 @@ description: Options for the Python convert function
 | `hybrid_mode`             | `str`                | `"auto"`     | Hybrid triage mode. Values: auto (default, dynamic triage), full (skip triage, all pages to backend)                               |
 | `hybrid_url`              | `str`                | -            | Hybrid backend server URL (overrides default)                                                                                      |
 | `hybrid_timeout`          | `str`                | `"30000"`    | Hybrid backend request timeout in milliseconds. Default: 30000                                                                     |
-| `hybrid_fallback`         | `bool`               | `True`       | Fallback to Java processing on hybrid backend error. Default: true                                                                 |
+| `hybrid_fallback`         | `bool`               | `False`      | Opt in to Java fallback on hybrid backend error (default: disabled)                                                                |
--- a/content/docs/cli-options-reference.mdx
+++ b/content/docs/cli-options-reference.mdx
@@ -36,7 +36,7 @@ This page documents all available CLI options for opendataloader-pdf.
 | `--hybrid-mode`             | -     | `string`  | `"auto"`     | Hybrid triage mode. Values: auto (default, dynamic triage), full (skip triage, all pages to backend)                               |
 | `--hybrid-url`              | -     | `string`  | -            | Hybrid backend server URL (overrides default)                                                                                      |
 | `--hybrid-timeout`          | -     | `string`  | `"30000"`    | Hybrid backend request timeout in milliseconds. Default: 30000                                                                     |
-| `--hybrid-fallback`         | -     | `boolean` | `true`       | Fallback to Java processing on hybrid backend error. Default: true                                                                 |
+| `--hybrid-fallback`         | -     | `boolean` | `false`      | Opt in to Java fallback on hybrid backend error (default: disabled)                                                                |
 
 ## Examples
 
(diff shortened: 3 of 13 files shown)


bundolee:
Thanks for the review! Here's how each issue was addressed:

**Issue 1 (processor-level integration test): Not adding — here's why.**
`HybridDocumentProcessor` is a static utility class that requires real PDF parsing and server connectivity. The fail-fast path has two layers of coverage:
- `DoclingFastServerClient.checkAvailability()` — tested with MockWebServer for both connection-refused and HTTP 503 cases
- `HybridDocumentProcessor.processDocument()` calls `getClient(config).checkAvailability()` as Phase 0 (line 112), before any triage — this is a direct, unconditional call with no branching that could silently swallow the exception

A processor-level test would require mocking the entire PDF pipeline (content extraction, page filtering) just to reach the `checkAvailability()` call, adding complexity without meaningful additional coverage. The wiring is a single line with no conditional logic.

**Issue 2 (flag description): Fixed.**
Updated description from `"Fallback to Java processing on hybrid backend error. Default: false"` to `"Opt in to Java fallback on hybrid backend error (default: disabled)"`. Propagated via `npm run sync`.

**Issue 3 (server.shutdown() side-effect): Already handled.**
`@BeforeEach` creates a fresh `MockWebServer` instance and calls `server.start()` for every test. `@AfterEach` calls both `client.shutdown()` and `server.shutdown()`. Each test gets an isolated server lifecycle — no shared state between tests.

**Issue 4 (public-facing docs): Acknowledged.** The docs changes are auto-generated from `options.json` and reflect the correct new defaults.

Pushed the Issue 2 fix.

bundolee:
Thanks for the continued review.

**Issue 1 (double-shutdown):** MockWebServer.shutdown() is idempotent — calling it twice is safe and does not produce errors or noisy stack traces. Isolating a single test into a @Nested class with its own lifecycle adds structural complexity without meaningful benefit. Keeping it as-is.

**Issue 2 (CHANGELOG):** This project auto-generates the CHANGELOG from conventional commit messages at release time. The commit message (`fix: change hybrid_fallback default to false`) will appear in the next release notes automatically. No manual entry needed.

**Issue 3 (Python bool design debt):** Agreed this is pre-existing and out of scope for this PR.

**Nit (test rename):** Good point — will rename `testHybridConfigFallback` → `testHybridConfigFallbackToggle` for clarity.

bundolee:
Addressed all actionable feedback across review rounds. Remaining items are either already covered or out of scope:

- **Double-shutdown**: As noted previously, MockWebServer.shutdown() is idempotent. No behavioral risk.
- **SDK migration note**: The PR description already documents the breaking change and migration path. SDK callers who omit `hybridFallback` get the new default — this is the intended behavior change.
- **Error message assertion style**: Current approach is clear and maintainable for the scope of these tests.

This PR is ready for merge. All tests pass, benchmarks are neutral, and `npm run sync` has been run.

PR #288, opened 2026-03-16
fix: propagate processing failures as non-zero exit code in CLIMain

### LINKED ISSUE #287: Hybrid backend failure exits with code 0 when fallback is disabled
## Summary

When `--hybrid docling-fast` is set with `--hybrid-fallback` disabled (default), and the hybrid backend server is unreachable, the Java process logs the `IOException` at SEVERE level but still exits with code 0. This prevents the Python wrapper (`subprocess.run(check=True)`) from detecting the failure.

## Current Behavior

```bash
opendataloader-pdf --hybrid docling-fast --hybrid-url http://127.0.0.1:59999 file.pdf
# SEVERE: IOException connecting to hybrid backend
# Exit code: 0 (no output files produced)
```

## Expected Behavior

- `hybrid-fallback=false` (default) + server unreachable → **exit code != 0**
- `hybrid-fallback=true` + server unreachable → exit code 0 + Java-only extraction result

## Impact

`langchain-opendataloader-pdf` relies on non-zero exit codes to propagate errors to users. With exit code 0, hybrid failures are silently swallowed and users get zero documents with no feedback.

## Reproduction

```python
import opendataloader_pdf

opendataloader_pdf.convert(
    input_path=["test.pdf"],
    output_dir="output/",
    hybrid="docling-fast",
    hybrid_url="http://127.0.0.1:59999",  # unreachable
    hybrid_fallback=False,
)
# Expected: exception raised
# Actual: no exception, no output files
```

## Related

- `langchain-opendataloader-pdf` hybrid mode integration depends on this fix
- Integration tests `test_hybrid_fallback_on_bad_url` and `test_hybrid_no_fallback_on_bad_url` are blocked

bundolee:
## Summary
Fixes #287

- `CLIMain.processFile()` caught all exceptions and logged them at SEVERE level but never propagated failure to the exit code — the JVM always exited with 0
- Refactored `CLIMain` to expose a testable `run()` method that returns an exit code
- `processFile()`, `processPath()`, and `processDirectory()` now return boolean success/failure indicators

## Problem
When `--hybrid docling-fast` targets an unreachable server with `--hybrid-fallback` disabled (default), the Java process logs the `IOException` at SEVERE level but exits with code 0. This prevents `langchain-opendataloader-pdf`'s `subprocess.run(check=True)` from detecting the failure, silently swallowing hybrid errors.

## Changes
- `java/.../cli/CLIMain.java` — Extracted `run(String[])` returning exit code; changed `processFile`/`processPath`/`processDirectory` from `void` to `boolean`; `main()` calls `System.exit(exitCode)` on failure
- `java/.../cli/CLIMainTest.java` — New test class with 3 tests: processing failure returns non-zero, directory with failing file returns non-zero, no arguments returns zero

## Approach
Minimal refactor that changes return types from `void` to `boolean` to propagate failure up the call chain. No new dependencies, no behavioral changes for successful processing.

## How to test
```bash
cd java && mvn test -pl opendataloader-pdf-cli -Dtest=CLIMainTest -Dsurefire.failIfNoSpecifiedTests=false -am
```

- Before fix: `CLIMain.run(["--hybrid", "docling-fast", "--hybrid-url", "http://127.0.0.1:59999", "test.pdf"])` → exit code 0
- After fix: same invocation → exit code 1

## Breaking changes
None. Successful processing still exits with code 0. Only failure cases now correctly exit with code 1.

### DIFF
--- a/java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIMain.java
+++ b/java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIMain.java
@@ -54,7 +54,7 @@ public class CLIMain {
         } catch (ParseException e) {
             System.out.println(e.getMessage());
             formatter.printHelp(HELP, options);
-            return 0;
+            return 2;
         }
 
         // Handle --export-options before requiring input files
@@ -77,7 +77,7 @@ public class CLIMain {
         } catch (IllegalArgumentException exception) {
             System.out.println(exception.getMessage());
             formatter.printHelp(HELP, options);
-            return 0;
+            return 2;
         }
         configureLogging(quiet);
         boolean hasFailure = false;
@@ -112,7 +112,7 @@ public class CLIMain {
     private static boolean processPath(File file, Config config) {
         if (!file.exists()) {
             LOGGER.log(Level.WARNING, "File or folder " + file.getAbsolutePath() + " not found.");
-            return true;
+            return false;
         }
         if (file.isDirectory()) {
             return processDirectory(file, config);
@@ -126,7 +126,7 @@ public class CLIMain {
         File[] children = file.listFiles();
         if (children == null) {
             LOGGER.log(Level.WARNING, "Unable to read folder " + file.getAbsolutePath());
-            return true;
+            return false;
         }
         boolean allSucceeded = true;
         for (File child : children) {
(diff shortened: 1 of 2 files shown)


bundolee:
Thanks for the thorough review! Here's how the feedback was addressed:

**Already fixed in follow-up commit (ca182dc):**
- `ParseException` / `IllegalArgumentException` now return exit code **2** (POSIX convention)
- Non-existent files and unreadable directories now return `false` (failure)
- Added tests for both cases (`testInvalidArgumentsReturnsExitCode2`, `testNonExistentFileReturnsNonZeroExitCode`)

**Not adopting:**
- `--hybrid-timeout 1000` in tests: the malformed PDF triggers a parse failure before any network call is made, so the unreachable URL never causes a timeout in practice.
- `tempDir.resolve` for non-existent path: `File.exists()` returns `false` for any non-existent path regardless of OS — no platform-specific behavior involved.

All 54 tests passing, no CI timeout risk.

PR #296, opened 2026-03-18
fix(test): add StreamInfo to test TextChunks for verapdf 1.31.8 compat

bundolee:
## Summary
- verapdf `wcag-algorithms` 1.31.8 added `streamInfos` field to `BaseObject` and changed `TextChunk.getTextChunk()` to call `StreamInfo.updateStreamInfos()`, which crashes with `IndexOutOfBoundsException` when `streamInfos` is empty
- Production TextChunks from the PDF parser always have `streamInfos` populated, but test-created TextChunks did not — 4 tests were failing:
  - `TableBorderProcessorTest.testProcessTableBorders`
  - `TableBorderProcessorTest.testProcessTableBordersDepthLimitNoStackOverflow`
  - `ContentSanitizerTest.testMultipleReplacementsInSingleChunk`
  - `ContentSanitizerTest.testReplaceCoveringPartsOfChunks`
- Added `StreamInfo` initialization to test TextChunk creation to match production behavior
- Also proactively fixed `TableBorderProcessorTest.testProcessTableBordersNormalNestedTableProcessedCorrectly` which uses the same code path but was not yet failing (it would break once its TextChunk hits a multi-cell split)

> Note: `StreamInfo.updateStreamInfos()` not guarding against empty lists is an upstream bug. Filed as [veraPDF/veraPDF-wcag-algs#375](https://github.com/veraPDF/veraPDF-wcag-algs/issues/375).

## Test plan
- [x] `mvn clean test` — all 418 tests pass (364 core + 54 cli)
- [x] No production code changes, only test fixtures updated

### DIFF
--- a/java/opendataloader-pdf-core/src/test/java/org/opendataloader/pdf/processors/TableBorderProcessorTest.java
+++ b/java/opendataloader-pdf-core/src/test/java/org/opendataloader/pdf/processors/TableBorderProcessorTest.java
@@ -28,7 +28,6 @@ import org.verapdf.wcag.algorithms.entities.tables.tableBorders.TableBorder;
 import org.verapdf.wcag.algorithms.entities.tables.tableBorders.TableBorderCell;
 import org.verapdf.wcag.algorithms.entities.tables.tableBorders.TableBorderRow;
 import org.verapdf.wcag.algorithms.semanticalgorithms.containers.StaticContainers;
-
 import org.verapdf.wcag.algorithms.semanticalgorithms.utils.StreamInfo;
 
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


bundolee:
Addressed feedback 1 and 2 in `bf51946`:

1. **`null` xObjectName** — `StreamInfo.updateStreamInfos()` never accesses `xObjectName`, and `null` is the correct value for test TextChunks that aren't backed by a real PDF stream. Added a comment explaining this at the first call site.

2. **Import ordering** — Moved `StreamInfo` import into the `org.verapdf` block, removed stray blank line.

3. **Helper abstraction in `TableBorderProcessorTest`** — Not adopting this. Each test creates TextChunks with different parameters (bbox, text, font sizes) specific to their scenario. A shared factory would either need many parameters (not simpler) or obscure test intent. The StreamInfo line is a one-liner that's clear in context.

bundolee:
PR description updated — added `testProcessTableBordersNormalNestedTableProcessedCorrectly` as a proactive fix. It wasn't failing yet but uses the same `TextChunk.getTextChunk` code path and would break once a multi-cell split triggers `StreamInfo.updateStreamInfos`.

PR #300, opened 2026-03-19
fix: filter narrow outlier elements in vertical gap detection

### LINKED ISSUE #294: Issues with Text Ordering (XYCut) and Borderless Table Extraction
Hi,

I am currently using opendataloader-pdf v2.0.0 and have observed some issues regarding the text ordering and table extraction in certain PDFs.

1. Text Ordering Issues (XYCut vs. Off)
While XYCut is generally effective for multi-column layouts, I found cases where the left-to-right order is swapped:

- With XYCut: Pages 63 and 568 show swapped left/right column order.

- With XYCut Off: Page 568 is ordered correctly, but Page 63 still shows swapped order.

Additionally, even within a single column (the left side), the sequence of sentences is sometimes inconsistent.

Could you clarify if these issues stem from the internal structure of the PDF itself or if they are edge cases that the library’s sorting logic hasn't covered yet?

2. Extraction of Borderless Tables
I also tested a PDF containing tables without vertical/horizontal lines using Hybrid Mode. However, the results were similar to the standard mode, and the table structure was not captured correctly. Is this a known limitation for borderless tables, or are there specific configurations I should adjust?

I have attached the relevant PDF for your review. I would appreciate it if you could look into these cases.

bundolee:
## Summary
Fixes #294

Narrow elements (page numbers, footnote markers) bridging the column gap reduce the edge vertical gap below `MIN_GAP_THRESHOLD` (5pt), preventing column boundary detection in two-column layouts.

## Problem
The XY-Cut++ vertical gap detection scans left-to-right edges to find the largest gap. When a narrow element (e.g., 16pt wide page number) sits in the ~20pt column gap, the largest edge gap drops to ~2pt (left column edge → bridge left, and bridge right → right column edge), which is below the 5pt threshold. The algorithm then falls through to `sortByYThenX`, interleaving left and right column text.

## Changes
- `XYCutPlusPlusSorter.java` — When edge gap < 5pt, retry gap detection after filtering out elements narrower than 10% of region width (`NARROW_ELEMENT_WIDTH_RATIO`). Extracted `findVerticalCutByEdges` as a reusable private method.
- `XYCutPlusPlusSorterTest.java` — New test `sort_twoColumnsWithNarrowBridge_leftColumnFirst` verifying column detection with a narrow bridge element.

## Approach
Narrow outlier filtering was chosen over center-based column detection because it is simpler, lower risk, and only activates when the edge gap is already below threshold. When `edgeCut.gap >= 5pt`, the new code path is never entered — existing behavior is preserved exactly.

## How to test
```bash
cd java && mvn test -pl opendataloader-pdf-core -Dtest="XYCutPlusPlusSorterTest#sort_twoColumnsWithNarrowBridge_leftColumnFirst"
```

- Before fix (main): `AssertionFailedError: Left column should come before right column`
- After fix (this branch): PASS

## Benchmark
NID=0.9119 — identical to main, no regression across all 200 documents.

## Breaking changes
None. The filter only activates when edge gap < 5pt and objects >= 4.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/readingorder/XYCutPlusPlusSorter.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/readingorder/XYCutPlusPlusSorter.java
@@ -440,7 +440,7 @@ public class XYCutPlusPlusSorter {
         // When edge gap is small, narrow outlier elements (e.g., page numbers,
         // footnote markers) may bridge an otherwise clear column gap.
         // Retry without elements narrower than 10% of the region width.
-        if (objects.size() >= 4) {
+        if (objects.size() >= 3) {
             BoundingBox region = calculateBoundingRegion(objects);
             if (region != null) {
                 double regionWidth = region.getRightX() - region.getLeftX();
@@ -455,7 +455,7 @@ public class XYCutPlusPlusSorter {
                 }
                 if (filtered.size() >= 2 && filtered.size() < objects.size()) {
                     CutInfo filteredCut = findVerticalCutByEdges(filtered);
-                    if (filteredCut.gap > edgeCut.gap) {
+                    if (filteredCut.gap > edgeCut.gap && filteredCut.gap >= MIN_GAP_THRESHOLD) {
                         return filteredCut;
                     }
                 }


bundolee:
Addressed code review feedback in 41449e4:

1. **Lowered minimum object count from `>= 4` to `>= 3`** — A 3-element set (1 left column + 1 right column + 1 narrow bridge) is a valid real-world scenario. The inner guard `filtered.size() >= 2` already ensures meaningful processing after filtering.

2. **Added `MIN_GAP_THRESHOLD` check on filtered cut return** — Previously, if `edgeCut.gap = 2pt` and `filteredCut.gap = 3pt`, the method would return a sub-threshold result. Now the filtered cut is only returned when it both improves on the edge cut *and* meets the 5pt threshold. This makes the method's contract clearer — callers can trust that a non-zero gap result is actionable.

The SonarQube Cognitive Complexity warning (S3776) on `findBestVerticalCutWithProjection` is pre-existing and out of scope for this PR.

bundolee:
Responding to the second code review:

**1. `findVerticalCutByEdges` size guard**
Already covered — `findBestVerticalCutWithProjection` has an early return at the top:
```java
if (objects.size() < 2) {
    return new CutInfo(0, 0);
}
```
So `findVerticalCutByEdges` is never called with fewer than 2 elements.

**2. -0.1% NID/MHS regression**
This is a stale baseline issue, not a real regression. The baseline was not updated before this PR was opened, so the comparison is against an older snapshot. The scores are identical when compared against the actual current main output.

**3. 3-element boundary test**
Acknowledged as a valid coverage gap. Will track as a follow-up.

PR #320, opened 2026-03-20
fix: handle null textColor in HeadingProcessor for hybrid mode

justperson94:
## Summary

Fixes #319

- Add defensive try-catch for NullPointerException in `HeadingProcessor.processHeadings()` and `HeadingProcessor.detectHeadingsLevels()` when verapdf's `getTextColor()` returns null for hybrid backend-generated text nodes.
- Follows the same defensive pattern used in #134 and #220 for similar verapdf NPEs.

## Changes

**File**: `opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/HeadingProcessor.java`

1. `processHeadings()`: Wrap `NodeUtils.headingProbability()` call in try-catch; skip the node on NPE
2. `detectHeadingsLevels()`: Wrap `TextStyle.getTextStyle()` call in try-catch; skip the heading on NPE
3. Add `Logger` for FINE-level diagnostic logging

## Test plan

Tested with 4 PDF files in hybrid mode (`--hybrid docling-fast`):

| File | Pages | Before fix | After fix |
|------|-------|-----------|-----------|
| Simple resume | 1 | OK | OK |
| Complex tables document | 92 | **NPE, no output** | OK |
| Image-heavy document | 58 | OK | OK |
| PPT-to-PDF portfolio | 21 | **NPE, no output** | OK |

Non-hybrid (fast) mode also tested — no regression.

Issue #319, opened 2026-03-20
NullPointerException in HeadingProcessor during hybrid mode postProcess (textColor null)


justperson94:
### Bug

When using hybrid mode (`--hybrid docling-fast`), the Java CLI crashes with a NullPointerException during post-processing. The docling-fast backend processes pages successfully (returns HTTP 200), but the Java process fails in `HeadingProcessor` because text nodes from the hybrid backend lack color information.

This affects two code paths:

1. `HeadingProcessor.processHeadings()` → `NodeUtils.headingProbability()` → `SemanticTextNode.getTextColor()` → `calculateTextColor()` → `Arrays.stream(null)`
2. `HeadingProcessor.detectHeadingsLevels()` → `TextStyle.getTextStyle()` → `SemanticTextNode.getTextColor()` → `calculateTextColor()` → `Arrays.stream(null)`

The crash causes the entire file to fail with no output generated. `hybrid_fallback=True` does not help because the error occurs in post-processing (after the backend succeeds), not in the backend call itself.

### Stack trace (path 1)

```
java.lang.NullPointerException: Cannot read the array length because "array" is null
    at java.base/java.util.Arrays.stream(Arrays.java:5619)
    at org.verapdf.wcag.algorithms.entities.SemanticTextNode.calculateTextColor(SemanticTextNode.java:390)
    at org.verapdf.wcag.algorithms.entities.SemanticTextNode.getTextColor(SemanticTextNode.java:379)
    at org.verapdf.wcag.algorithms.semanticalgorithms.utils.NodeUtils.headingProbability(NodeUtils.java:116)
    at org.opendataloader.pdf.processors.HeadingProcessor.processHeadings(HeadingProcessor.java:74)
    at org.opendataloader.pdf.processors.HeaderFooterProcessor.processHeaderOrFooterContent(HeaderFooterProcessor.java:144)
    at org.opendataloader.pdf.processors.HybridDocumentProcessor.postProcess(HybridDocumentProcessor.java:488)
```

### Stack trace (path 2)

```
java.lang.NullPointerException: Cannot read the array length because "array" is null
    at java.base/java.util.Arrays.stream(Arrays.java:5619)
    at org.verapdf.wcag.algorithms.entities.SemanticTextNode.calculateTextColor(SemanticTextNode.java:390)
    at org.verapdf.wcag.algorithms.entities.SemanticTextNode.getTextColor(SemanticTextNode.java:379)
[...]

PR #350, opened 2026-03-24
fix: replace PR #320 defensive NPE catches with proper graceful degradation

bundolee:
## Problem

PR #320 (#319 fix) added broad try-catch blocks around `getTextColor()` call sites to prevent NPE from verapdf's `calculateTextColor()` in hybrid mode. While the crash was real, the defensive approach caused three behavioral regressions:

1. **Heading detection silently disabled for hybrid pages** — `safeGetTextColor()==null → continue` skipped all heading probability calculation, producing zero headings (MHS metric regression).

2. **Text over-fragmentation** — `oneLineProbability = 0` fallback in TextLineProcessor forced every text chunk into a separate line, splitting paragraphs that should be merged.

3. **Headings with no level assigned** — `detectHeadingsLevels()` caught NPE and skipped the heading entirely, leaving it with default level 0 instead of a meaningful heading level.

Additionally, `HeadingProcessor.safeGetTextColor()` was imported by SerializerUtil (serialization layer), creating an inverted dependency.

## Solution

**Graceful degradation** instead of skip-on-null:

- `TextNodeUtils.getTextColorOrDefault()` — returns black `{0,0,0}` instead of null, so heading detection and debug strings keep working.
- `TextNodeUtils.getTextColorOrNull()` — returns null for serialization, where omitting the field is preferred over fabricating a value.
- `HeadingProcessor.detectHeadingsLevels()` — colorless headings get level assigned by font-size proximity to existing style groups, instead of being dropped from the level map.
- `HeadingProcessor.processHeadings()` — null-color nodes no longer skipped; they participate in heading probability calculation normally.
- `TextLineProcessor` — NPE catch removed entirely; `countOneLineProbability()` does not call `getTextColor()` (verified via bytecode analysis).

**Architecture**: `safeGetTextColor()` removed from HeadingProcessor; shared utility in `org.opendataloader.pdf.utils.TextNodeUtils` eliminates the inverted dependency from serializers to processors.

### 1. Static verification (bytecode analysis of wcag-algorithms 1.31.11)

- `calculateTextColor()`: confirmed `ifnull` guard at offset 117 before `Arrays.stream()` — null `getFontColor()` is skipped, not streamed.
- `NodeUtils.headingProbability()`: calls `getTextColor()` at offset 102 — defended by TextNodeUtils in our code and by verapdf's fix upstream.
- `TextStyle.getTextStyle()`: calls `getTextColor()` at offset 19, passes to constructor. `TextStyle.compareTo()` only compares maxFontSize and fontWeight (not textColor), so null color is safe in TreeMap ordering.
- `ChunksMergeUtils.countOneLineProbability()`: does NOT call `getTextColor()` — uses position, font size, and baseline only. The TextLineProcessor NPE catch was unnecessary from the start.

### 2. Unit tests

- `mvn test`: 455 tests, 0 failures, 0 errors

### 3. Benchmark (200 PDFs, Java-only mode)

| | NID | TEDS | MHS |
|---|---|---|---|
| Baseline (v1.6.2) | 0.9126 | 0.4942 | 0.6588 |
| This PR (v1.9.1) | 0.9131 | 0.4942 | **0.6715** |
| Delta | +0.0005 | +0.0000 | **+0.0126** |

No regression. MHS (heading structure) improved +1.26% — consistent with the fix restoring heading detection that PR #320 had suppressed.

### 4. Hybrid mode integration test (docling-fast backend)

Tested 5 PDFs with hybrid mode (docling-fast server on port 5002):
- All 5 documents processed without NPE crash
- Headings detected and level assigned (level=1) in all documents
- JSON serialization correctly omits textColor when unavailable (via `getTextColorOrNull`)

Fixes issues raised in PR #320 review.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Refactor**
  * Improved text color retrieval with centralized utility and safer error handling.
  * Enhanced heading level detection for PDFs with colorless headings using font-size proximity matching.
  * Simplified text processing logic by removing redundant error handling checks.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/json/serializers/SerializerUtil.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/json/serializers/SerializerUtil.java
@@ -17,7 +17,7 @@ package org.opendataloader.pdf.json.serializers;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import org.opendataloader.pdf.json.JsonName;
-import org.opendataloader.pdf.processors.HeadingProcessor;
+import org.opendataloader.pdf.utils.TextNodeUtils;
 import org.verapdf.wcag.algorithms.entities.IObject;
 import org.verapdf.wcag.algorithms.entities.SemanticTextNode;
 
@@ -46,7 +46,7 @@ public class SerializerUtil {
     public static void writeTextInfo(JsonGenerator jsonGenerator, SemanticTextNode textNode) throws IOException {
         jsonGenerator.writeStringField(JsonName.FONT_TYPE, textNode.getFontName());
         jsonGenerator.writePOJOField(JsonName.FONT_SIZE, textNode.getFontSize());
-        double[] textColor = HeadingProcessor.safeGetTextColor(textNode);
+        double[] textColor = TextNodeUtils.getTextColorOrNull(textNode);
         if (textColor != null) {
             jsonGenerator.writeStringField(JsonName.TEXT_COLOR, Arrays.toString(textColor));
         }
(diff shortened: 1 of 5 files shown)


bundolee:
@coderabbitai full review

PR #373, opened 2026-03-31
fix(node): add shims: true to tsup config to fix CJS build

debuggerpk:
## Problem

When the package is consumed via `require()` (CJS), the following error is thrown:

```
TypeError: The "path" argument must be of type string or an instance of URL. Received undefined
```

The root cause is in the compiled `dist/index.cjs`. tsup emits:

```js
var import_meta = {};
var __filename = fileURLToPath(import_meta.url); // import_meta.url is undefined
```

`import.meta.url` is ESM-only. When tsup compiles to CJS without shims enabled, it replaces `import.meta` with an empty object `{}`, so `.url` is `undefined` and `fileURLToPath(undefined)` throws.

## Fix

Add `shims: true` to `tsup.config.ts`.

This causes tsup to inject a proper CJS-compatible shim in the `.cjs` output:

```js
var import_meta = { url: require("url").pathToFileURL(__filename).href };
```

which makes `fileURLToPath(import_meta.url)` resolve correctly.

## Changes

- `node/opendataloader-pdf/tsup.config.ts`: add `shims: true`

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to optimize bundling and output generation processes.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/entities/SemanticPicture.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/entities/SemanticPicture.java
@@ -81,4 +81,40 @@ public class SemanticPicture extends BaseObject {
     public boolean hasDescription() {
         return description != null && !description.isEmpty();
     }
+
+    /**
+     * Returns a sanitized version of the description safe for use as alt text
+     * across all output formats (Markdown, HTML, JSON) without format-specific escaping.
+     *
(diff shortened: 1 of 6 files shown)


debuggerpk:
thanks for the package guys. however this becomes a problem on commonjs based systems. 
@bundolee can you please review?

bundolee (approved):
Thanks for the fix and the well-written explanation! `shims: true` is the correct solution for CJS compatibility with `import.meta.url`.

LGTM, approving.

PR #375, opened 2026-03-31
Fix: handle invalid input paths in Java API to prevent crash

pavanpai769:
###  Problem

When using the Java API (`OpenDataLoaderPDF.processFile`) in a loop (as shown in the documentation https://opendataloader.org/docs/quick-start-java ), providing an invalid file path causes an exception and stops the entire processing.

###  Reproduction Example

The issue can be reproduced using the documented example:

```java
import org.opendataloader.pdf.api.Config;
import org.opendataloader.pdf.api.OpenDataLoaderPDF;

public class Sample {
    public static void main(String[] args) throws Exception {
        Config config = new Config();
        config.setOutputFolder("path/to/output");
        config.setGeneratePDF(true);
        config.setGenerateMarkdown(true);
        config.setGenerateHtml(true);

        try {
            // Process multiple files in one JVM invocation
            for (String pdf : new String[]{"valid.pdf", "invalid.pdf"}) {
                OpenDataLoaderPDF.processFile(pdf, config);
            }
        } finally {
            OpenDataLoaderPDF.shutdown();
        }
    }
}

```

### ❌ Current Behavior

If one of the input paths is invalid or does not exist, the program throws an exception and stops execution, preventing other valid files from being processed.

### ⚠️ Inconsistency

- CLI handles invalid paths gracefully  
- Java API does not  
- Documentation suggests batch processing should work reliably  

### ✅ Fix

- Added input file validation in `OpenDataLoaderPDF.processFile`
- If a path is invalid:
  - Log a warning
  - Skip processing that file
- Continue processing remaining valid inputs

### 📈 Impact

- Prevents crashes during batch processing  
- Improves robustness of the Java API  
- Aligns behavior with CLI and documented usage  

### 🧪 Testing

Tested manually with:
- Valid paths only  
- Mix of valid and invalid paths  
- All invalid paths  

Unit tests can be added if required.

## Summary by CodeRabbit

* **Bug Fixes**
  - PDF inputs are validated before processing: null/blank, invalid paths, missing or non-regular files, and non-.pdf names are detected; invalid inputs are skipped and validation failures are logged.

* **Changes**
  - Processing now returns a result object indicating success or failure instead of throwing exceptions; failures are logged and reported in the returned result.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/api/OpenDataLoaderPDF.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/api/OpenDataLoaderPDF.java
@@ -19,12 +19,19 @@ import org.opendataloader.pdf.hybrid.HybridClientFactory;
 import org.opendataloader.pdf.processors.DocumentProcessor;
 
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * The main entry point for the opendataloader-pdf library.
  * Use the static method {@link #processFile(String, Config)} to process a PDF.
  */
 public final class OpenDataLoaderPDF {
+
+    private static final Logger LOGGER = Logger.getLogger(OpenDataLoaderPDF.class.getCanonicalName());
 
     private OpenDataLoaderPDF() {
     }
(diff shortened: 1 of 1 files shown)


bundolee (changes requested):
Thanks for identifying this gap between CLI and Java API behavior — it's a real usability issue for batch processing.

However, the current approach silently swallows all errors (including real `IOException` from corrupt PDFs, disk-full, etc.) by catching and logging them. This removes the caller's ability to detect failures, which is a breaking change for existing users who rely on the `throws IOException` contract.

A library API should report errors — the caller decides how to handle them. The documented batch loop example can simply wrap `processFile` in a try-catch:

```java
for (String pdf : files) {
    try {
        OpenDataLoaderPDF.processFile(pdf, config);
    } catch (IOException e) {
        logger.warning("Skipping " + pdf + ": " + e.getMessage());
    }
}
```

If you'd like to add input validation (null/blank/invalid path), that's reasonable — but it should throw `IllegalArgumentException` so the caller can catch it, not silently swallow it. The `throws IOException` on the signature should stay.

Also a minor typo: `"File not fount at"` → `"File not found at"`.

Would you be open to reworking this so validation throws rather than swallows?

bundolee:
Closing this PR as there's been no response to the review feedback for 2+ weeks and CI is still failing.

The underlying problem is valid — I've opened #430 to track it. The key design constraint: the API should throw, not swallow errors. See the review thread above for details.

@pavanpai769 Thanks for identifying this issue! If you'd like to take another shot based on the feedback, feel free to reference #430.

PR #406, opened 2026-04-10
fix: update CID warning hybrid guidance

### LINKED ISSUE #404: Warning suggests --hybrid-mode for OCR fallback, but hybrid backend is not enabled without --hybrid
## Summary

When OpenDataLoader detects CID-keyed fonts without ToUnicode mappings, it logs a warning that says:

> Text extraction may be incomplete. Consider using `--hybrid-mode` for OCR fallback.

This is misleading because `--hybrid-mode` alone does not enable the hybrid backend. According to the CLI options and docs, users must also pass `--hybrid <backend>` (for example `--hybrid docling-fast`) for OCR/backend processing to happen at all.

## Current Behavior

The warning in `ContentFilterProcessor` currently says:

```java
"Text extraction may be incomplete. Consider using --hybrid-mode for OCR fallback."
```

Location:

- `java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/ContentFilterProcessor.java`

## Why this is a problem

`--hybrid-mode` is only a mode selector (`auto` / `full`), not the switch that enables the hybrid backend.

From the CLI/docs/generated bindings, hybrid usage looks like:

```bash
opendataloader-pdf --hybrid docling-fast file.pdf
```

and optionally:

```bash
opendataloader-pdf --hybrid docling-fast --hybrid-mode full file.pdf
```

So the current warning points users to an option that is insufficient on its own.

### Warning text
- `java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/ContentFilterProcessor.java`

### CLI option semantics
- `java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIOptions.java`
- `content/docs/cli-options-reference.mdx`
- `README.md`
- generated Python/Node bindings (`convert_generated.py`, `convert-options.generated.ts`)

### Tests
`CLIOptionsTest` covers `--hybrid-mode` together with `--hybrid`, which matches the documented behavior.

## Suggested Fix

Update the warning to recommend a complete actionable command, for example:

- `Consider using --hybrid docling-fast for OCR fallback.`
- or, if needed, `Consider using --hybrid docling-fast --hybrid-mode full ...`

A more explicit message might be:

> Text extraction may be incomplete. Consider enabling hybrid OCR fallback with `--hybrid docling-fast`.

## Expected Outcome

Users seeing this warning should be directed to the actual option that enables backend/OCR processing, not only the sub-mode selector.

LocNguyenSGU:
## Summary

Update the CID-font replacement-character warning to recommend the current hybrid OCR command instead of the obsolete `--hybrid-mode` flag.

## Reproduction

`ContentFilterProcessor` warns when a page is mostly replacement characters (U+FFFD), but the warning text still pointed users to `--hybrid-mode for OCR fallback`.

## Patch Summary

- change the warning text to recommend `--hybrid docling-fast`
- tighten `CidFontDetectionTest` so it asserts the new guidance and rejects the stale wording

## Risks

Low. This is a message-only correction plus a focused test update.

## Validation

- `cd java && mvn -pl opendataloader-pdf-core -Dtest=CidFontDetectionTest test`

Resolves #404

## Summary by CodeRabbit

* **Documentation**
  * Updated warning message to recommend the correct command-line flag for enabling hybrid OCR fallback when replacement character issues are detected.

* **Tests**
  * Updated test to validate the improved warning message guidance.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/ContentFilterProcessor.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/ContentFilterProcessor.java
@@ -78,7 +78,7 @@ public class ContentFilterProcessor {
             LOGGER.log(Level.WARNING,
                 "Page {0}: {1,number,#.#%} of characters are replacement characters (U+FFFD). "
                 + "This PDF likely contains CID-keyed fonts without ToUnicode mappings. "
-                + "Text extraction may be incomplete. Consider using --hybrid-mode for OCR fallback.",
+                + "Text extraction may be incomplete. Consider enabling hybrid OCR fallback with --hybrid docling-fast.",
                 new Object[]{pageNumber + 1, replacementCharRatio});
         }
         TextProcessor.replaceUndefinedCharacters(pageContents, config.getReplaceInvalidChars());
--- a/java/opendataloader-pdf-core/src/test/java/org/opendataloader/pdf/processors/CidFontDetectionTest.java
+++ b/java/opendataloader-pdf-core/src/test/java/org/opendataloader/pdf/processors/CidFontDetectionTest.java
@@ -120,9 +120,11 @@ public class CidFontDetectionTest {
             );
 
             boolean hasReplacementWarning = warnings.stream()
-                .anyMatch(w -> w.contains("replacement characters"));
+                .anyMatch(w -> w.contains("replacement characters")
+                    && w.contains("--hybrid docling-fast")
+                    && !w.contains("--hybrid-mode for OCR fallback"));
             Assertions.assertTrue(hasReplacementWarning,
-                "Expected WARNING log about replacement characters");
+                "Expected WARNING log about replacement characters with actionable hybrid guidance");
         } finally {
             logger.removeHandler(handler);
         }


LocNguyenSGU:
Thanks — I checked this more closely.

The CodeRabbit “Docstring Coverage” item appears to be a generic repository-level warning in the comment summary, not a failing GitHub status for this PR. The actual PR checks are currently green (`CodeRabbit`, `license/cla`), and this patch only updates an existing warning string plus its focused test assertion.

Since this change does not introduce new public APIs/functions, I did not add unrelated docstrings just to satisfy a repo-wide coverage warning that is outside the scope of issue #404.

If maintainers want, I can still add a separate follow-up PR for broader docstring coverage, but I’d prefer to keep this bug-fix PR focused on the warning guidance change.

bundolee (approved):
Good catch — the old warning pointed users to `--hybrid-mode`, which is just a mode selector and doesn't actually enable the hybrid backend. This fix gives them the correct, actionable flag.

Clean change, focused test update. Thanks!

One thought for a future issue: recommending hybrid OCR as the go-to fix for CID font mapping problems may not always be the best advice — OCR can actually be worse than incomplete text extraction on vector-text PDFs. But that's a separate discussion about the diagnostic message itself, not this PR's scope.

Issue #405, opened 2026-04-10
Image paths with spaces break markdown preview


MischaD:
### Bug

When the input PDF filename contains spaces, the extracted markdown references
images via a relative path that inherits that filename — also containing spaces.
Most markdown renderers stop parsing the path at the first space, so images fail
to load silently.

### Steps to reproduce

1. Run `opendataloader_pdf.convert()` on a PDF whose filename contains spaces,
   e.g. `my paper.pdf`
2. Open the generated `.md` file in any markdown renderer (VS Code, GitHub, etc.)
3. Images are not displayed.

**Root cause:** the generated image links use an unescaped path, e.g.:

`![image 1](my paper_images/imageFile1.png)`

Unescaped spaces terminate the link destination per the CommonMark spec.

**Suggested fix:** URL-encode spaces in the image path when writing markdown output:

`![image 1](my%20paper_images/imageFile1.png)`

Alternatively, replace spaces with underscores in the `_images/` subdirectory name at creation time so the path never needs escaping.

### Version

2.2.1

### Java version

java version "18.0.1" 2022-04-19
Java(TM) SE Runtime Environment (build 18.0.1+10)
Java HotSpot(TM) 64-Bit Server VM (build 18.0.1+10, mixed mode, sharing)

ankitsingh0913:
Hey @MischaD,
I’ve identified and fixed the issue where images extracted from PDFs fail to render when filenames contain spaces.

The fix ensures directory names are normalized by replacing spaces with underscores before being passed to StaticLayoutContainers, preventing broken paths (e.g., my_paper_images instead of my paper_images).

The project builds cleanly after this change.
Happy to open a PR or take this up if you’d like to assign it to me.

ankitsingh0913:
I've done it, raising PR

bundolee:
Hi @MischaD — thanks for the careful repro and the spec-aware suggested fix.

Fixed in #468 (commit 4b87a30, on `main` now).

**What landed**

Image link destinations are now wrapped in CommonMark §6.4 angle-bracket form (`<...>`), which is the spec-recommended way to embed paths containing spaces, parens, brackets, etc. The on-disk image directory stays byte-identical to the rendered link in the `.md`, so reading the link tells you exactly where the file is.

For your reproducer, the output goes from this:
```markdown
![image 1](my paper_images/imageFile1.png)   <!-- broken: renderer truncates at the space -->
```
to this:
```markdown
[...]

PR #411, opened 2026-04-11
Bug Fix

ankitsingh0913:
I have addressed the issue where images extracted from PDFs with filenames containing spaces fail to render in Markdown format.

Following your suggestion, I updated DocumentProcessor.java to replace spaces with underscores in the image directory name before it is assigned to StaticLayoutContainers.

This ensures both the output directory and referenced paths use consistent naming (for example, my_paper_images instead of my paper_images), which resolves the broken links without requiring separate URL encoding logic for Markdown or HTML generators.

I also verified that the project compiles successfully after the change.

Please let me know if any additional changes or testing are needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Image output directory names now preserve filename stems (handles names without extensions) and replace spaces with underscores to avoid malformed image folders when no custom image directory is set.

* **Chores**
  * Adjusted a dependency version range for improved resolution and compatibility.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/DocumentProcessor.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/DocumentProcessor.java
@@ -325,7 +325,7 @@ public class DocumentProcessor {
                 imagesDirectory = config.getImageDir();
             } else {
                 String fileName = Paths.get(inputPdfName).getFileName().toString();
-                String baseName = fileName.substring(0, fileName.length() - 4);
+                String baseName = fileName.substring(0, fileName.length() - 4).replace(" ", "_");
                 imagesDirectory = config.getOutputFolder() + File.separator + baseName + MarkdownSyntax.IMAGES_DIRECTORY_SUFFIX;
             }
             StaticLayoutContainers.setImagesDirectory(imagesDirectory);
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/DocumentProcessor.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/DocumentProcessor.java
@@ -326,6 +326,9 @@ public class DocumentProcessor {
             } else {
                 String fileName = Paths.get(inputPdfName).getFileName().toString();
                 String baseName = fileName.substring(0, fileName.length() - 4).replace(" ", "_");
++               int dotIndex = fileName.lastIndexOf('.');
++               String rawBaseName = dotIndex > 0 ? fileName.substring(0, dotIndex) : fileName;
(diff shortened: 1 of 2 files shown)


bundolee:

   - Others → @bundolee

bundolee:
Hi @ankitsingh0913 — thank you for tackling this! Apologies for the slow turnaround.

We ended up landing a different fix for #405 in #468 (commit 4b87a30), which I want to walk through so the reasoning is on the record.

**What your PR did**

Replaced spaces (and you mentioned a broader sanitize in the description) with underscores in the on-disk image directory name, so `my paper.pdf` produces a `my_paper_images/` directory and the Markdown link `(my_paper_images/imageFile1.png)` parses cleanly.

**Why we went a different way**

The on-disk-sanitize approach has two issues that surfaced during review:

1. **`--image-dir` reproduces the same bug class.** A user explicitly setting `--image-dir "my pictures"` would still land an unsafe path inside the rendered link. Sanitizing that path silently mangles explicit user input, which violates the CLI principle that user-given paths are kept verbatim. So the fix would either need to silently rewrite the user's directory (surprising) or stay narrow and leave the same bug open via the custom-dir door.

2. **CommonMark already specifies a form for this case.** Section 6.4 defines an angle-bracket link-destination form (`<...>`) precisely so paths containing spaces, parens, brackets, etc. can be embedded without rewriting. That's what we adopted: the on-disk path stays byte-identical to the rendered link, and the user-supplied filename or `--image-dir` value is preserved everywhere.

```markdown
![image 1](<my paper_images/imageFile1.png>)
```

**Other things from your PR worth keeping**

You also replaced the hard-coded `length() - 4` (which assumes a `.pdf` extension) with `lastIndexOf('.')`. That's a real improvement orthogonal to #405 — it correctly handles uppercase `.PDF`, multiple-dot filenames like `report.v1.2.pdf`, and other extensions. **Could you split that out into a separate small PR?** Happy to merge it once the implementation has tests.

Thanks again for the contribution and for stepping up on a real bug. Closing this PR in favor of #468, but please don't take that as a "your work wasn't useful" signal — the design discussion this PR triggered is what got us to the spec-correct fix. Looking forward to the `lastIndexOf('.')` follow-up if you want to send it.

bundolee:
Closed in favor of #468 (commit 4b87a30). See preceding comment for the design rationale. Thanks again, @ankitsingh0913.

ankitsingh0913:
Hi @bundolee , thanks a lot for the detailed review and for taking the time to explain the reasoning — really appreciate it, especially the clarification around the Common Mark approach and the CLI design principles.

That makes complete sense, and I’m glad the discussion contributed to a better, spec-aligned solution.

I’ll go ahead and raise a separate PR for the lastIndexOf('.') improvement along with proper tests.

Thanks again for the guidance!

PR #420, opened 2026-04-13
await code

tejasae-afk:
Was reading through node/opendataloader-pdf/src/cli.ts and the promise result gets used as if it were already resolved. this patch adds the missing await so the async path is actually sequenced. Not sure if this is intentional, so feel free to ignore.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.

## Summary by CodeRabbit

* **Refactor**
  * Improved CLI startup sequence for more reliable asynchronous operation handling.

### DIFF
--- a/node/opendataloader-pdf/src/cli.ts
+++ b/node/opendataloader-pdf/src/cli.ts
@@ -71,7 +71,7 @@ async function main(): Promise<number> {
   }
 }
 
-main().then((code) => {
+await main().then((code) => {
   if (code !== 0) {
     process.exit(code);
   }


bundolee:
Thanks for looking into this — appreciate the careful read-through of the codebase.

The existing `main().then(...)` pattern is intentional here. The `.then()` callback already runs after `main()` resolves, so adding `await` doesn't change the sequencing. Since this is a CLI entry point (not imported as a module), the top-level await form would behave identically but adds a subtle semantic difference (blocking module evaluation) that we'd rather avoid for no gain.

No changes needed, but thanks for flagging it!

Issue #430, opened 2026-04-15
Java API: invalid file path crashes batch processing (CLI handles gracefully)


bundolee:
## Problem

When using `OpenDataLoaderPDF.processFile` in a loop, an invalid file path throws an exception and stops the entire batch. The CLI handles this gracefully, but the Java API does not.

Originally reported in #375 by @pavanpai769.

## Reproduction

```java
for (String pdf : new String[]{"valid.pdf", "invalid.pdf"}) {
    OpenDataLoaderPDF.processFile(pdf, config);
}
```

If `invalid.pdf` doesn't exist, the exception kills the loop and `valid.pdf` (if after) never gets processed.

## Expected behavior

The API should throw a clear exception (e.g. `IllegalArgumentException` for invalid paths) so callers can catch and skip, rather than silently swallowing errors. The `throws IOException` contract must be preserved.

## Design direction

- Add input validation (null/blank/non-existent/non-PDF) that throws `IllegalArgumentException`
- Keep `throws IOException` on the signature — callers decide error handling
- Document the batch pattern with try-catch in Javadoc

See review discussion on #375 for details.

pavanpai769:
@bundolee I’ve opened PR #431 for this—would appreciate your review.

PR #431, opened 2026-04-15
Fix: Handle invalid file paths in OpenDataLoaderPDF

pavanpai769:
Fixes the issue discussed in #375.

### Changes:
- Added input validation for invalid file paths (null, blank, non-existent, non-PDF)
- Throws IllegalArgumentException for invalid inputs
- Preserved existing IOException contract

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened pre-processing validation for input files (blank/invalid paths, non-existent or non-regular files, and non-.pdf names), with clearer error reporting and warning-level logging to improve feedback on invalid inputs.

* **Documentation**
  * Updated method documentation to reflect the new validation behavior and removed an outdated exception declaration.


bundolee:

   - Others → @bundolee

hnc-jglee:
Hi @pavanpai769 — thank you for tackling #430 and for the clear write-up here. The direction was right, and we've reworked it into a follow-up PR (#564). Sharing the reasoning so it's transparent:

**What changed from this PR's approach**
1. **Exception type** — invalid input now throws the existing `InvalidPdfFileException` (a checked subtype of `IOException`) instead of `IllegalArgumentException`. The issue asks to *preserve the `throws IOException` contract* and behave *like the CLI in batch*; an unchecked `IllegalArgumentException` would bypass the CLI's existing `catch (InvalidPdfFileException)` skip-and-continue path, so a single checked type keeps batch loops working.
2. **Validation location** — moved into `DocumentProcessor.preprocessing()` next to the existing `validatePdfMagicNumber()` guard, rather than the `OpenDataLoaderPDF` facade. This avoids duplicating the file-existence checks and a time-of-check/time-of-use race.
3. **Path disclosure** (the point CodeRabbit raised) — log messages now use the file name only instead of the full absolute path, and the log-then-throw was dropped so callers control logging.
4. **Tests** — added `Issue430IntegrationTest` with the batch regression that proves the fix: a valid PDF placed *after* a bad path still gets processed.

None of this is a knock on your work — it surfaced the right problem and the validation cases you enumerated all carried over. Really appreciate the contribution! 🙏

Issue #400, opened 2026-04-10
Error: std::bad_alloc persists in v2.2.1


i-Veni-Vidi-Vici:
### Bug

`std::bad_alloc` occurs consistently on pages 23–30 of a 30-page PDF during the C++ preprocess stage in docling-fast backend. This reproduces on **every run including run 1** after a fresh server restart, ruling out memory accumulation.

Two additional issues identified:
**Lost page numbers**: Server logs show pages 23–30 failed individually, but the client only receives `"Page 30: std::bad_alloc"` + 8x `"Unknown page: pipeline terminated early"`, **preventing accurate fallback routing and causing page loss**.

**Server-side log (docling-fast):**

2026-04-10 11:21:11,965 - INFO - detected formats: [<InputFormat.PDF: 'pdf'>]
2026-04-10 11:21:11,965 - INFO - Going to convert document batch...
2026-04-10 11:21:11,965 - INFO - Processing document tmpc3stf0fp.pdf
2026-04-10 11:21:14,501 - ERROR - Stage preprocess failed for run 2, pages [23]: std::bad_alloc
2026-04-10 11:21:14,516 - ERROR - Stage preprocess failed for run 2, pages [24]: std::bad_alloc
2026-04-10 11:21:14,535 - ERROR - Stage preprocess failed for run 2, pages [25]: std::bad_alloc
2026-04-10 11:21:14,550 - ERROR - Stage preprocess failed for run 2, pages [26]: std::bad_alloc
2026-04-10 11:21:14,565 - ERROR - Stage preprocess failed for run 2, pages [27]: std::bad_alloc
2026-04-10 11:21:14,584 - ERROR - Stage preprocess failed for run 2, pages [28]: std::bad_alloc
2026-04-10 11:21:14,599 - ERROR - Stage preprocess failed for run 2, pages [29]: std::bad_alloc
2026-04-10 11:21:14,619 - ERROR - Stage preprocess failed for run 2, pages [30]: std::bad_alloc
 UserWarning: 'pin_memory' argument is set as true but no accelerator is found, then device pinned memory won't be used.
  super().__init__(loader)
2026-04-10 11:21:31,535 - INFO - Finished converting document tmpc3stf0fp.pdf in 19.58 sec.
2026-04-10 11:21:31,551 - WARNING - Docling returned partial_success: 8 error(s), failed_pages will be reported

client log

[...]

PR #434, opened 2026-04-16
Fix: detect failed pages via error messages and gap detection union

i-Veni-Vidi-Vici:
## Summary

- Add `_extract_failed_pages_from_errors()` to parse failed page numbers from Docling error messages (`"Page N: <error>"`)
- Change `build_conversion_response()` to **union** both error-message parsing and gap detection strategies instead of using fallback
- Docling may include failed pages as empty entries in the pages dict, making gap-only detection miss them. Combining both strategies ensures neither failure mode is lost.

## Test plan

- [x] `_extract_failed_pages_from_errors()` unit tests (std::bad_alloc, mixed formats, no-match)
- [x] Integration test: both strategies combined — gap + error parsing returns union `[2, 4, 5]`
- [x] Integration test: duplicate page errors deduplicated
- [x] Integration test: no "Page N:" pattern falls back to gap detection
- [x] Integration test: empty errors list still detects gaps
- [x] All 48 existing tests pass (no regression)
- [x] Benchmark: NID 0.902, TEDS 0.489, MHS 0.740 — no regression

## Related Issue
- https://github.com/opendataloader-project/opendataloader-pdf/issues/400

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failed-page detection during PDF conversion by combining error-message parsing of "Page N:" entries with missing-page gap detection; failures are deduplicated, sorted, and correctly reported even when page entries are empty or the top-level pages key is missing.

* **Tests**
  * Added tests covering error-string page extraction, gap detection, deduplication, boundary cases, combined failure reporting, and fallback behavior when error messages lack page markers.


bundolee:

   - Others → @bundolee

bundolee (approved):

bundolee (approved):
LGTM — verified the diagnosis and the fix. Thanks for the thorough investigation and the clean patch.

**What I verified locally:**
- Reproduced the union semantics: `_extract_failed_pages_from_errors()` correctly parses `"Page N: <msg>"` from docling's error format (matches `docling/pipeline/standard_pdf_pipeline.py:720-725`), and `build_conversion_response()` unions error-parsing with gap-detection.
- Java side has no additional bug: `HybridResponse.hasFailedPages()` is purely `!failedPages.isEmpty()`, and the fallback gate at `HybridDocumentProcessor.java:254` (`if (!backendFailedPages.isEmpty())`) only checks emptiness. So the Python-only patch is sufficient to unblock the fallback that was silently failing on the reporter's run.
- Ran `pytest tests/test_hybrid_server_partial_success.py` against this branch — 23/23 pass.
- No regression on existing tests; `test_partial_success_missing_pages_key` still expects `[1, 2, 3]` (only its docstring was updated for accuracy).

**Optional follow-ups (not blocking):**

1. **Regex misses the no-colon variant.** Docling's source emits `f"{page_label}: {error_msg}" if error_msg else page_label` — when `error_msg` is empty, the message is bare `"Page 26"` with no colon. The current regex `r"^Page\s+(\d+):"` would miss that. Suggest making the colon optional: `r"^Page\s+(\d+)(?::|$)"`. Doesn't affect the `std::bad_alloc` case being fixed here, but strictly more robust.
2. If you take #1, please add a test like `assert _extract_failed_pages_from_errors(["Page 26"]) == [26]`.

Both are 5-minute changes and welcome as a follow-up commit, but I don't want to block this on them — the fix as-is solves the reported bug.

Approving for squash-merge.

PR #467, opened 2026-04-30
fix(hybrid): match bare "Page N" in failed-page error parser

i-Veni-Vidi-Vici:
Relax `_extract_failed_pages_from_errors` regex from `^Page\s+(\d+):` to `^Page\s+(\d+)(?::|$)` so the bare "Page N" form (no colon) that docling's `StandardPdfPipeline` emits when `error_msg` is empty is also captured.

Follow-up to #434 implementing the two optional suggestions in @bundolee's review:
1. Relax the regex to make the colon optional.
2. Add a test for the bare "Page 26" form.

Defensive change: in the wrapper's current code path `failed_pages` always carry a non-empty `RuntimeError` message, so the practical occurrence is rare, but the one-character cost protects against future docling/wrapper changes.

**Verification:**
- `pytest tests/test_hybrid_server_partial_success.py` — 24/24 passing (existing 23 + new `test_bare_page_no_colon`)
- `./scripts/bench.sh --check-regression` — all thresholds met

**Checklist:**


i-Veni-Vidi-Vici:
Not adopting the CodeRabbit suggestion on `hybrid_server.py:133-138`.

Verified against docling 2.88.0 source — the variants the bot proposes (`"Page 26 "` with trailing space, `"Page 26  : <error>"` with spaces before the colon) don't exist in docling's emitted formats. `standard_pdf_pipeline.py:763` produces `"Page N: <msg>"` (no space before colon) or bare `"Page N"` (no trailing whitespace); `base_pipeline.py:348` and `vlm_pipeline.py:235,247` likewise emit without extra whitespace. Loosening the regex would only match phantom inputs.

bundolee (approved):
Thanks for the clean follow-up @i-Veni-Vidi-Vici 🙏

**Objective:** Capture the bare `"Page N"` form that docling's `StandardPdfPipeline` can emit when `error_msg` is empty — a defensive widening so `_extract_failed_pages_from_errors` stays robust against future docling/wrapper drift.

**Approach review:** Regex change `^Page\s+(\d+):` → `^Page\s+(\d+)(?::|$)` is the minimal possible diff. I verified empirically that:
- `"Page 26: std::bad_alloc"` still matches (no regression on the existing path)
- `"Page 26"` now matches (new bare form)
- `"Page 26abc"`, `"Page 26 something"`, `"Page 26 - error"` correctly do **not** match — boundary stays tight

Docstring update is accurate, scope is disciplined (exactly the two hunks promised), and the new `test_bare_page_no_colon` locks in the new behavior.

**Evidence:** 24/24 partial-success tests + bench regression check both green per PR description.

**Optional follow-up (non-blocking, please don't gate on this):** consider a future PR adding a negative test like `_extract_failed_pages_from_errors(["Page 26abc", "Page 26 something"]) == []` to lock in the boundary intent and block accidental loosening later. Truly optional.

LGTM — approving and rebase-merging. Thanks again!

PR #468, opened 2026-04-30
fix(markdown): wrap image link destinations in angle brackets to handle unsafe filename chars

### LINKED ISSUE #405: Image paths with spaces break markdown preview

bundolee:
A user processing `my paper.pdf` (or any filename with spaces, parens, brackets, etc.) gets a Markdown file where every image link silently fails to render in VS Code, GitHub, and similar viewers. The same class of bug also reproduces through `--image-dir "my pictures"`, where the user's explicit path is well-formed on disk but breaks in the rendered Markdown.

Fixes [opendataloader-project/opendataloader-pdf#405](https://github.com/opendataloader-project/opendataloader-pdf/issues/405)

## Approach
Wrap the relative path in the CommonMark angle-bracket link-destination form (`<...>`) at the Markdown emit site (`writeImage` / `writePicture` in `MarkdownGenerator`), instead of rewriting the on-disk directory name.

Why this layer and this form:
- **Angle-bracket is the spec-recommended way.** CommonMark §6.4 explicitly admits the angle-bracket form for destinations containing spaces, parens, brackets, etc. The bare form `(...)` would terminate at the first reserved character. Only `<`, `>`, and `\` need escaping inside `<...>`, and we backslash-escape those; newlines have no representable form and are replaced with spaces.
- **The on-disk path stays byte-identical to the rendered link.** This preserves user intent for both the default `<stem>_images/` directory and any `--image-dir` value, and lets users locate their images on disk by reading the link.
- **JSON `source` fields are left raw.** Downstream consumers treat that field as a filesystem path; encoding or wrapping would break their lookups. Different sublanguage, different escape rules.
- **HTML output is left unchanged.** Quoted attribute values legally accept spaces/parens/brackets and the user agent handles `src` URL encoding when the page is rendered. Verified end-to-end with a local HTTP server.

## Evidence
Built the CLI locally and ran the full repro path on three scenarios. Also re-ran the full core test suite (566 tests) to confirm no regression.

| `my paper.pdf` (default dir, before fix) | Image link contains literal space → renderer truncates | `image 1` — broken in VS Code/GitHub |
| `my paper.pdf` (default dir, after fix) | On-disk dir verbatim; MD link in angle brackets | On-disk `my paper_images/`, MD link `image 1` ✅ |
| `draft (v2) [rc].pdf` (after fix) | Parens & brackets verbatim inside `<...>` | MD link `image 1 [rc]_images/imageFile1.png>)` ✅ |
| `--image-dir "my pictures (v2)"` (after fix) | User's path verbatim on disk; MD link in angle brackets | On-disk `my pictures (v2)/`, MD link `<my pictures (v2)/imageFile1.png>` ✅ |
| JSON `source` field (after fix) | Raw filesystem path (no wrapping) | `"source" : "my paper_images/imageFile1.png"` ✅ |
| `mvn -pl opendataloader-pdf-core -am test` | All pass | 566 tests, 0 failures, BUILD SUCCESS ✅ |

Unit tests in `MarkdownGeneratorTest` cover the helper directly: safe paths still get wrapped (uniform output), spaces / parens / brackets / apostrophes / quotes / literal `%` / `#` all pass through verbatim inside `<...>`, literal `<` / `>` / `\` in the path get backslash-escaped, newlines collapse to spaces, non-ASCII passes through, null returns null. Integration tests in `ImageDirIntegrationTest` exercise the full pipeline end-to-end for both default and `--image-dir` paths with reserved characters.

## Notes
- Supersedes #411, which took the alternative approach of rewriting the on-disk directory name. That approach silently mutates explicit user input via `--image-dir` (a CLI hygiene violation) and reproduces the same bug class through the custom-dir door.
- An earlier revision of this PR percent-encoded the link instead of using the angle-bracket form. The angle-bracket form is the CommonMark spec's intended idiom for this case, keeps the on-disk path byte-identical to the link, and avoids cross-format encoding considerations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Markdown image link destinations are now formatted safely: special characters (including <, >, backslashes), spaces and newlines are normalized/escaped and wrapped in angle brackets so links render reliably.

* **Tests**
  * Added integration and unit tests validating Markdown image-link formatting for paths with reserved characters and newline variants.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/markdown/MarkdownGenerator.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/markdown/MarkdownGenerator.java
@@ -150,6 +150,39 @@ public class MarkdownGenerator implements Closeable {
         }
     }
 
+    /**
+     * Wraps an image relative path as a CommonMark angle-bracket link destination
+     * (`<...>`). The bare form `(my paper.png)` is terminated by the first space or
+     * unbalanced parenthesis, so paths inheriting filenames with spaces, parens, or
+     * brackets break in renderers (#405). The angle-bracket form is the
+     * spec-recommended way to embed such paths and lets the on-disk path stay
+     * byte-identical to the rendered link, which preserves user intent for both the
+     * default `<stem>_images/` directory and any `--image-dir` value.
+     *
+     * Only `<`, `>`, and `\` are reserved inside the angle-bracket form; escape
+     * those with a backslash. Newlines have no representable form in a link
+     * destination — replace them with spaces so the destination stays well-formed.
+     */
+    static String formatMarkdownLinkDestination(String path) {
+        if (path == null) {
+            return null;
+        }
+        StringBuilder sb = new StringBuilder(path.length() + 2);
+        sb.append('<');
+        for (int i = 0; i < path.length(); i++) {
+            char c = path.charAt(i);
+            if (c == '<' || c == '>' || c == '\\') {
+                sb.append('\\').append(c);
+            } else if (c == '\n' || c == '\r') {
+                sb.append(' ');
+            } else {
+                sb.append(c);
+            }
+        }
+        sb.append('>');
+        return sb.toString();
+    }
+
     protected void writeImage(ImageChunk image) {
         try {
             String absolutePath = String.format(MarkdownSyntax.IMAGE_FILE_NAME_FORMAT, StaticLayoutContainers.getImagesDirectory(), File.separator, image.getIndex(), imageFormat);
@@ -164,7 +197,7 @@ public class MarkdownGenerator implements Closeable {
                         LOGGER.log(Level.WARNING, "Failed to convert image to Base64: {0}", absolutePath);
                     }
                 } else {
-                    imageSource = relativePath;
+                    imageSource = formatMarkdownLinkDestination(relativePath);
                 }
                 if (imageSource != null) {
                     String altText = (image instanceof EnrichedImageChunk && ((EnrichedImageChunk) image).hasDescription())
@@ -198,7 +231,7 @@ public class MarkdownGenerator implements Closeable {
                         LOGGER.log(Level.WARNING, "Failed to convert image to Base64: {0}", absolutePath);
                     }
                 } else {
-                    imageSource = relativePath;
+                    imageSource = formatMarkdownLinkDestination(relativePath);
                 }
                 if (imageSource != null) {
                     String altText = picture.hasDescription()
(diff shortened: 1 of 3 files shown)


bundolee:
Responding to @coderabbitai's three review comments. All three make the same recommendation — replace the angle-bracket form with percent-encoding of the path contents. **I'm declining the recommendation; it's based on a misreading of the CommonMark spec.**

The chosen fix uses CommonMark §6.4's angle-bracket link-destination form (`<...>`). Per spec:

> A link destination consists of either … a sequence of zero or more characters between an opening `<` and a closing `>` that contains no line endings or unescaped `<` or `>` characters

Inside `<...>`, the only reserved characters are `<`, `>`, `\`, and line endings. Spaces, parens, brackets, apostrophes, percent signs, and hashes are **explicitly admissible** without encoding. Percent-encoding the path inside `<...>` is not required by the spec, would *change* the path that the renderer resolves (`%20` is decoded back to space by the renderer's URL handling, but `<my paper.png>` is parsed verbatim as the literal path `my paper.png`), and breaks byte-identity between the on-disk path and the rendered link.

Verified compatibility:
- CommonMark reference parser (cmark / cmark-gfm — what GitHub uses): supports.
- markdown-it (what VS Code uses): supports.
- pandoc: supports.
- marked: supports.

Verified during review:
- 32/32 unit + integration tests pass with the angle-bracket form.
- End-to-end smoke test with `my paper.pdf`, `draft (v2) [rc].pdf`, and `--image-dir "my pictures (v2)"` — all three produce links that GitHub-flavored renderers parse correctly with the on-disk path byte-identical to the link.

If a downstream user reports a renderer that supports percent-encoded but not angle-bracket destinations (very unlikely — angle-bracket form is older and more universally supported), the fix would be to add an opt-in flag, not to revert. Keeping spec-correct, byte-identical paths as the default.

Leaving the merge as-is. Thanks for the careful read, @coderabbitai — but in this case the simpler form really is the spec-recommended one.

PR #473, opened 2026-04-30
fix(node): split CLI streaming from library API to prevent stdout double-write

### LINKED ISSUE #398: fix(node): stdout output is written twice when quiet option is not set — streamOutput in executeJar conflicts with CLI output layer
## Bug Description

In `node/opendataloader-pdf/src/index.ts`, when `convert()` is called from the CLI without the `quiet` flag, output is written to `process.stdout` **twice**.

### Root Cause

`executeJar()` is called with `streamOutput: !options.quiet`. When `streamOutput=true`, every data chunk from the Java process is immediately written to `process.stdout`:

```typescript
javaProcess.stdout.on('data', (data) => {
  const chunk = data.toString();
  if (streamOutput) {
    process.stdout.write(chunk);  // <-- writes to stdout in real-time
  }
  stdout += chunk;  // <-- also accumulated
});
```

Then the resolved `stdout` string is returned back to `cli.ts`, which writes it again:

```typescript
// cli.ts
const output = await convert(inputPaths, convertOptions);
if (output && !convertOptions.quiet) {
  process.stdout.write(output);  // <-- second write of the same content!
}
```

### Fix

When `streamOutput=true`, the accumulated `stdout` variable should be returned as an empty string (since it was already written to `process.stdout\' in real-time), preventing the CLI from double-printing. 

```typescript
javaProcess.on('close', (code) => {
  if (code === 0) {
    // If already streamed, return empty to avoid double-write by caller
    resolve(streamOutput ? '' : stdout);
  } else {
    ...
  }
});
```

### Impact

All CLI users who do not pass `--quiet` see every line of output duplicated. This is a significant UX regression.

### Affected File

`node/opendataloader-pdf/src/index.ts`

bundolee:
## Objective
CLI users running `opendataloader-pdf <file> --to-stdout` (or any invocation
without `--quiet`) see every output line printed twice. The cause: `executeJar()`
streams Java's stdout to `process.stdout` in real time *and* accumulates it
into the resolved string, then `cli.ts` prints that string a second time.

This PR supersedes #399 by @kuishou68, whose diagnosis was correct and whose
original fix is preserved as the basis of this change. We took it over because
(a) #399's approach made `convert()` (the library API) regress to returning an
empty string when `quiet` was undefined — flagged in CodeRabbit's second review
and unaddressed for ~12 days — and (b) we wanted to land regression coverage at
the same time so this class of bug stops recurring.

Fixes [opendataloader-project/opendataloader-pdf#398](https://github.com/opendataloader-project/opendataloader-pdf/issues/398)
Closes [opendataloader-project/opendataloader-pdf#399](https://github.com/opendataloader-project/opendataloader-pdf/pull/399)

## Approach
Separate concerns at the layer boundary so that streaming (a CLI-only need) and
the resolved string (a library-only need) never compete for the same return
channel.

- `convert()` (library API) is pinned to `streamOutput: false`. It always
  returns the full stdout string and never touches the parent process's stdio.
  No `quiet`-dependent return shape, no surprise side-effects.
- `_runForCli()` (internal helper, used only by the bundled CLI) is pinned to
  `streamOutput: true`. It forwards Java's stdout and stderr to the parent
  process in real time and resolves to `void`, so the CLI cannot re-print what
  was already streamed.
- `cli.ts` calls `_runForCli()` and never touches the resolved value.

Java's progress logs continue to surface on stderr in real time — the property
that makes long hybrid-mode runs observable.

Test infrastructure was added because we kept rediscovering the same class of
bug by hand:
- `executeJar.unit.test.ts` (9 tests) — mocks `child_process.spawn` and asserts
  the streaming/buffering contract deterministically (no real Java needed).
- `streaming.integration.test.ts` (3 tests) — runs the bundled CLI as a real
  subprocess against a multi-page sample PDF and asserts the user-facing
  behavior: no double-write, stderr arrives before stdout completes, library
  call doesn't leak to the parent's stdio.

## Evidence
Manual verification on the bundled CLI plus full vitest run:

| CLI, no `--quiet`, `--to-stdout` (1-page PDF) | 459 bytes (single copy) | 459 bytes |
| CLI, no `--quiet`, `--to-stdout` (14-page PDF) | 63006 bytes (single copy) | 63006 bytes |
| 14-page CLI: stderr progress logs | streamed before stdout | 28 stderr lines, "Number of pages" present |
| Library: `convert(path, { quiet: true })` returns | full stdout string | 459 bytes |
| Library: `convert(path, { quiet: undefined })` returns | full stdout string | 459 bytes |
| Library: side-effect on `process.stdout` | none | none |
| Library: side-effect on `process.stderr` | none | none |
| `vitest --run` | all pass | 30/30 pass |

## Summary by CodeRabbit

* **Bug Fixes**
  * Eliminated CLI output duplication so conversion results no longer print twice.

* **Improvements**
  * Real-time forwarding of conversion stdout/stderr without duplicating payloads.
  * Library conversion now returns full subprocess output and avoids writing subprocess stdout to the parent process.
  * Improved UTF‑8 handling across chunked output and clearer exit/error reporting.
  * Build now strips internal declarations from emitted artifacts.

* **Tests**
  * Added unit and integration tests for streaming, interleaved stdout/stderr, and UTF‑8 correctness.

### DIFF
--- a/node/opendataloader-pdf/src/cli.ts
+++ b/node/opendataloader-pdf/src/cli.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import { Command, CommanderError } from 'commander';
-import { convert } from './index.js';
+import { _runForCli } from './index.js';
 import { CliOptions, buildConvertOptions } from './convert-options.generated.js';
 import { registerCliOptions } from './cli-options.generated.js';
 
@@ -56,13 +56,9 @@ async function main(): Promise<number> {
   const convertOptions = buildConvertOptions(cliOptions);
 
   try {
-    const output = await convert(inputPaths, convertOptions);
-    if (output && !convertOptions.quiet) {
-      process.stdout.write(output);
-      if (!output.endsWith('\n')) {
-        process.stdout.write('\n');
-      }
-    }
+    // _runForCli streams stdout/stderr to the parent process as they arrive;
+    // we deliberately do not re-print anything here. (Issue #398.)
+    await _runForCli(inputPaths, convertOptions);
     return 0;
   } catch (err) {
     const message = err instanceof Error ? err.message : String(err);
--- a/node/opendataloader-pdf/src/index.ts
+++ b/node/opendataloader-pdf/src/index.ts
@@ -1,6 +1,7 @@
 import { spawn } from 'child_process';
 import * as path from 'path';
 import * as fs from 'fs';
+import { StringDecoder } from 'string_decoder';
 import { fileURLToPath } from 'url';
 
 // Re-export types and utilities from auto-generated file
@@ -15,6 +16,11 @@ const __dirname = path.dirname(__filename);
 const JAR_NAME = 'opendataloader-pdf-cli.jar';
 
 interface JarExecutionOptions {
+  /**
+   * When true, forwards Java's stdout and stderr chunks to the parent
+   * process in real time as well as accumulating them. Used by the bundled
+   * CLI so long-running conversions show progress as it happens.
+   */
   streamOutput?: boolean;
 }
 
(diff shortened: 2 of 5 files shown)


bundolee:
Re: CodeRabbit's latest pass (Round 5) — declining the three open suggestions to keep this PR focused on #398. Notes for the record:

1. **Backpressure on `process.stdout.write`/`process.stderr.write`** (`src/index.ts`). Sound advice for a general production hardening pass, but unrelated to the double-write that #398 reports. `process.stdout`/`stderr` rarely apply backpressure in practice (TTY/pipe), and adding it changes the streaming semantics in ways that deserve their own change + tests.

2. **Fallback build using `pnpm run build` instead of an inline `tsup` invocation** (`test/streaming.integration.test.ts`). The inline form is intentional: the test only needs `dist/cli.js` and `dist/index.js`, so it skips `setup.cjs` (which depends on Maven artifacts being present) and DTS emission. Calling `pnpm run build` would add a Maven precondition to a test that should run from a Node-only checkout.

3. **`StringDecoder` in `captureSubprocess`** (`test/streaming.integration.test.ts`). The integration test runs against ASCII sample PDFs and asserts byte-equality against the JAR, so `Buffer.toString()` is correct here. Multi-byte UTF-8 boundary handling is already locked in by the unit suite (`executeJar.unit.test.ts`).

The runtime fix for #398 (CLI streaming/library API split) and its regression coverage are the contract this PR is meant to land. The points above are valid follow-ups; we'll handle them as separate work if and when they bite.

PR #483, opened 2026-05-04
fix(runner): use utf-8 encoding for subprocess I/O on Windows

hyunhee-jo:
## Objective
On Korean Windows (and other non-UTF-8 Windows locales), running `opendataloader-pdf` fails immediately with:
```
UnicodeDecodeError: 'cp949' codec can't decode byte 0xe2 in position 130: illegal multibyte sequence
```
The JAR outputs UTF-8, but the Python wrapper reads it using the system locale encoding (`cp949`).

## Approach
Replace `locale.getpreferredencoding(False)` with a hard-coded `"utf-8"` in both call sites:
- `subprocess.run` (quiet mode, line 29)
- `subprocess.Popen` (streaming mode, line 39)

The JAR always outputs UTF-8 regardless of OS locale, so tying the decoder to the system locale was always wrong. Also switch `sys.stdout.write(line)` to `sys.stdout.buffer.write(line.encode("utf-8", errors="replace"))` so the decoded text reaches the terminal correctly on Windows where stdout may also default to `cp949`.

This change is safe on Linux/macOS — those platforms already default to UTF-8, so the behavior is identical.

## Evidence
Verified source after patch:

| Check | Expected | Actual |
|-------|----------|--------|
| `encoding="utf-8"` occurrences | 2 (both call sites) | 2 |
| `locale.getpreferredencoding` occurrences | 0 (fully removed) | 0 |
| `stdout.buffer.write` occurrences | 1 | 1 |
| Before fix (cp949 Windows) | UnicodeDecodeError on byte 0xe2 | UnicodeDecodeError |
| After fix (cp949 Windows) | subprocess reads UTF-8 correctly | Correct (no decode error) |

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured consistent UTF-8 encoding for both captured and live-streamed command output, improving handling of special and non-ASCII characters.
  * Live output now preserves replacement behavior on encoding errors while continuing to stream data in real time and collect output for error reporting.

### DIFF
--- a/python/opendataloader-pdf/src/opendataloader_pdf/runner.py
+++ b/python/opendataloader-pdf/src/opendataloader_pdf/runner.py
@@ -26,7 +26,7 @@ def run_jar(args: List[str], quiet: bool = False) -> str:
                     capture_output=True,
                     text=True,
                     check=True,
-                    encoding=locale.getpreferredencoding(False),
+                    encoding="utf-8",
                 )
                 return result.stdout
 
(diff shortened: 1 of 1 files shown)


bundolee (approved):
CodeRabbit 피드백 반영 완료 (locale import 제거, flush 추가, errors=replace, hasattr guard) — 머지합니다.

PR #495, opened 2026-05-12
fix(python): deduplicate CLI failure output

hyunhee-jo:
## Objective

When the bundled JAR exits non-zero (e.g. `--pages -10`), the Python wrapper printed the JAR's error + help banner **three times** in a row. Reported internally as PDFDLOSP-4.

## Root cause

In [`runner.py`](python/opendataloader-pdf/src/opendataloader_pdf/runner.py) the streaming branch (`quiet=False`) writes the JAR's combined stdout/stderr to the console live as it arrives. The `except subprocess.CalledProcessError` handler then re-emitted the captured copy — twice, because `CalledProcessError.output` and `.stdout` are the **same attribute**, so `if error.output: print(Output: …)` followed by `if error.stdout: print(Stdout: …)` repeated the same buffer.

```
JAR streaming write       ← banner #1 (already on console)
Error running ...                                   ← meta
Return code: 2                                      ← meta
Output:  <captured copy>  ← banner #2 (duplicate of #1)
Stdout:  <captured copy>  ← banner #3 (alias of `output`)
```

## Approach

- Move the captured-stream prints behind a `quiet` guard. Only quiet mode has *not* surfaced the JAR output to the user, so it's the only path that should print the captured copy.
- Drop the legacy `Output:` line — `error.output` is an alias for `error.stdout`; printing both repeats the same text.
- Comment why the guard exists so the next reader doesn't undo it.

## Evidence

Reproducer:
```python
opendataloader_pdf.convert(input_path="samples/pdf/lorem.pdf", pages="-10", quiet=False)
```

| | before | after |
|---|---|---|
| total lines emitted | 515 | 196 |
| `Invalid page range format` occurrences | 3 | **1** |
| `usage:` banner occurrences | 3 | **1** |
| duplicate `Output:` / `Stdout:` pair | yes | **gone** |
| `Error running … / Return code: 2` | 1 | 1 (unchanged) |

Tests:
```
$ pytest python/opendataloader-pdf/tests/test_runner.py -v
test_streaming_failure_does_not_duplicate_output PASSED
test_quiet_failure_prints_captured_streams_once   PASSED
```

`tests/test_runner.py` covers both branches and is structured to fail loudly if either regression returns.

## Risk

- `quiet=True` path: no user-visible change — the captured stdout/stderr are still surfaced, just without the redundant `Output:` line. `Stdout:` / `Stderr:` headers preserved.
- `quiet=False` path: callers that grep wrapper stderr for the JAR's error text will need to grep wrapper stdout instead (where it was already being streamed). Documented behaviour was always streaming-to-console for the live branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate output when JAR execution encounters errors in non-quiet mode.
  * Improved error message handling in quiet mode to display captured output exactly once.

* **Tests**
  * Added regression tests for error output behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opendataloader-project/opendataloader-pdf/pull/495)

### DIFF
--- a/python/opendataloader-pdf/src/opendataloader_pdf/runner.py
+++ b/python/opendataloader-pdf/src/opendataloader_pdf/runner.py
@@ -78,10 +78,14 @@ def run_jar(args: List[str], quiet: bool = False) -> str:
     except subprocess.CalledProcessError as error:
         print("Error running opendataloader-pdf CLI.", file=sys.stderr)
         print(f"Return code: {error.returncode}", file=sys.stderr)
-        if error.output:
-            print(f"Output: {error.output}", file=sys.stderr)
-        if error.stderr:
-            print(f"Stderr: {error.stderr}", file=sys.stderr)
-        if error.stdout:
-            print(f"Stdout: {error.stdout}", file=sys.stderr)
+        # Streaming mode already wrote the JAR's output live to stdout, so
+        # re-printing the captured copy would duplicate it. Only surface the
+        # captured streams in quiet mode, where the caller has not seen them.
+        # Note: CalledProcessError.output and .stdout are aliases for the same
+        # attribute — printing both produces the same content twice.
+        if quiet:
+            if error.stdout:
+                print(f"Stdout: {error.stdout}", file=sys.stderr)
+            if error.stderr:
+                print(f"Stderr: {error.stderr}", file=sys.stderr)
         raise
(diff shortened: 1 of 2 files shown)


hnc-jglee (approved):
Thanks for the clean fix. The root-cause diagnosis is on point — verified locally that `CalledProcessError.output is CalledProcessError.stdout` (same attribute), and the streaming branch's `captured_output` is literally the live stream, so re-printing it on stderr was guaranteed duplication.

The quiet-mode guard is the right boundary, and the inline comment explaining *why* will save the next reader from undoing it. Tests cover both branches and would catch the regression on either side.

No core processor / renderer code touched, so skipping benchmark.

LGTM, approving.

PR #496, opened 2026-05-12
fix(cli): emit error for non-PDF top-level input

hyunhee-jo:
## Objective

A non-PDF file passed directly on the command line exits silently with code 0, leaving users unsure whether their file was processed. README declares PDF as the only supported input, so this silence is a UX bug rather than intended behavior.

## Approach

Introduce an `isTopLevel` flag in `processPath` to distinguish user-provided arguments from files discovered during directory traversal. Top-level non-PDF arguments now print an error message on stdout and make the run report failure via the existing `hasFailure` path. Files encountered while walking a directory continue to be skipped silently — this preserves the legitimate batch-folder use case (mixed PDF/non-PDF directories).

`npm run sync` is not required — no CLI option definitions changed. Python and Node wrappers pass Java stdout and exit code through unchanged.

## Evidence

Built before/after jars and ran the CLI against each:

| Scenario | Before | After |
|----------|--------|-------|
| top-level non-PDF (`abcd.png`) | silent, exit 0 | error message, exit 1 |
| folder with PDF + png | silent, exit 0 | silent, exit 0 (no regression) |
| mixed args (`doc.pdf abcd.png`) | png silent, exit 0 | png error, exit 1 |
| valid PDF only (sanity) | exit 0, `doc.json` | exit 0, `doc.json` (no regression) |

After-only stdout for top-level non-PDF:

```
Error: 'abcd.png' is not a PDF file. Input must be a PDF file or a folder containing PDF files.
```

Unit tests: `mvn -pl opendataloader-pdf-cli -am test -Dtest=CLIMainTest` → `Tests run: 10, Failures: 0, Errors: 0` (5 new regression-guard cases included: top-level non-PDF error message, top-level non-PDF exit code, directory silent-skip regression guard, directory-of-only-non-PDFs returning 0, mixed-argument partial-failure behavior).

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI now reports an error and fails for non-PDF top-level inputs, while silently skipping non-PDF files found inside directories; mixed top-level inputs fail overall with error reported for the non-PDF entry.

* **Tests**
  * Added regression tests covering top-level vs directory-child non-PDF handling, exit codes, and stdout behavior; includes helpers to capture and assert console output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opendataloader-project/opendataloader-pdf/pull/496)

### DIFF
--- a/java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIMain.java
+++ b/java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIMain.java
@@ -85,7 +85,7 @@ public class CLIMain {
         boolean hasFailure = false;
         try {
             for (String argument : arguments) {
-                if (!processPath(new File(argument), config)) {
+                if (!processPath(new File(argument), config, true)) {
                     hasFailure = true;
                 }
             }
@@ -110,15 +110,26 @@ public class CLIMain {
 
     /**
      * Processes a file or directory, returning true if all files succeeded.
+     *
+     * <p>{@code isTopLevel} distinguishes user-provided arguments from files
+     * discovered during directory traversal: a non-PDF given directly on the
+     * command line is reported as an error, while non-PDF files inside a
+     * directory are silently skipped (preserves batch-folder processing).
      */
-    private static boolean processPath(File file, Config config) {
+    private static boolean processPath(File file, Config config, boolean isTopLevel) {
         if (!file.exists()) {
             LOGGER.log(Level.WARNING, "File or folder " + file.getAbsolutePath() + " not found.");
             return false;
         }
         if (file.isDirectory()) {
             return processDirectory(file, config);
-        } else if (file.isFile()) {
+        }
+        if (file.isFile()) {
+            if (isTopLevel && !isPdfFile(file)) {
+                System.out.println("Error: '" + file.getName()
+                    + "' is not a PDF file. Input must be a PDF file or a folder containing PDF files.");
+                return false;
+            }
             return processFile(file, config);
         }
         return true;
@@ -132,7 +143,7 @@ public class CLIMain {
         }
         boolean allSucceeded = true;
         for (File child : children) {
-            if (!processPath(child, config)) {
+            if (!processPath(child, config, false)) {
                 allSucceeded = false;
             }
         }
--- a/java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIMain.java
+++ b/java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIMain.java
@@ -33,6 +33,8 @@ public class CLIMain {
 
     private static final String HELP = "[options] <INPUT FILE OR FOLDER>...\n Options:";
 
+    private enum InputSource { CLI_ARGUMENT, DIRECTORY_CHILD }
+
     public static void main(String[] args) {
         int exitCode = run(args);
         if (exitCode != 0) {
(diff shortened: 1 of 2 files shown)


hnc-jglee:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

hnc-jglee:
### Readability suggestion: replace `boolean isTopLevel` with a more self-documenting form

At the call sites in [`CLIMain.java#L85-L91`](https://github.com/opendataloader-project/opendataloader-pdf/blob/0af1a151b5e462920edaa75902a6722be50ebd8f/java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIMain.java#L85-L91) and [`#L143-L148`](https://github.com/opendataloader-project/opendataloader-pdf/blob/0af1a151b5e462920edaa75902a6722be50ebd8f/java/opendataloader-pdf-cli/src/main/java/org/opendataloader/pdf/cli/CLIMain.java#L143-L148), the bare `true` / `false` literal passed to `processPath` is a classic [boolean-parameter smell](https://martinfowler.com/bliki/FlagArgument.html) — the reader has to jump to the method signature to learn what the flag means:

```java
processPath(new File(argument), config, true);   // true = ?
processPath(child, config, false);                // false = ?
```

Three ways to fix, in increasing order of refactor cost:

**Option A — Named constant at the call site (smallest change)**
```java
private static final boolean TOP_LEVEL = true;
private static final boolean NESTED    = false;
...
processPath(new File(argument), config, TOP_LEVEL);
processPath(child, config, NESTED);
```
Pros: one-line change, no signature churn. Cons: still a boolean under the hood; doesn't scale if a third source ever appears.

**Option B — Enum for the input source (recommended)**
```java
private enum InputSource { CLI_ARGUMENT, DIRECTORY_CHILD }

private static boolean processPath(File file, Config config, InputSource source) {
    ...
    if (source == InputSource.CLI_ARGUMENT && !isPdfFile(file)) {
        System.out.println("Error: '" + file.getName() + "' is not a PDF file. ...");
        return false;
    }
    ...
}
```
Pros: call sites read like prose (`processPath(file, config, InputSource.CLI_ARGUMENT)`); extensible if new sources are added later (e.g., stdin, glob expansion); exhaustive `switch` becomes possible. Cons: tiny bit more code.

**Option C — Split into two methods (most explicit)**
```java
private static boolean processCliArgument(File file, Config config) {
    if (!file.exists()) { ... }
    if (file.isDirectory()) return processDirectory(file, config);
    if (file.isFile() && !isPdfFile(file)) {
        System.out.println("Error: '" + file.getName() + "' ...");
        return false;
    }
    return processFile(file, config);
}

private static boolean processDirectoryChild(File file, Config config) {
    // same, minus the non-PDF check
}
```
Pros: no flag at all; each method is single-purpose. Cons: some duplication between the two helpers; harder to keep in sync if `processPath` grows.

**My recommendation: Option B.** It removes the `true`/`false` ambiguity at the call site, scales cleanly (the codebase already shows signs of CLI input growth — `--to-stdout`, hybrid mode, etc.), and keeps the existence/directory checks in one place. Option A is a perfectly reasonable minimal fix if you'd prefer to keep this PR tightly scoped.

(Not blocking — the current code works correctly; this is a readability nit.)

hyunhee-jo:
Good catch — agreed, `isTopLevel` is a flag-argument smell.

Went with Option B (enum). Option A would still leave a boolean under the hood and Option C duplicates the existence/directory checks, while the enum makes the call sites read like prose (`processPath(file, config, InputSource.CLI_ARGUMENT)`) and keeps the door open for additional sources (stdin, glob, etc.) with no signature churn.

Pushed as `ae0a89b refactor(cli): replace isTopLevel flag with InputSource enum`. Pure refactor — verified by running the same before/after scenarios against the baseline (`0af1a15`) and new jars; all four scenarios produce identical output. Unit tests (`Tests run: 10`) pass unchanged.

hnc-jglee (approved):

PR #505, opened 2026-05-14
fix(tagged): assign ids to items materialised after struct-tree traversal

bundolee:
## Objective
When `--use-struct-tree` is enabled, JSON output omits `id` on most paragraphs and images while the same PDF without `--use-struct-tree` produces ids for every item. Downstream consumers that correlate items across passes (caption linking, hybrid backends, external pipelines) cannot match anything emitted via the tagged-PDF path.

Refs Jira PDFDLOSP-7.

## Approach
`TaggedDocumentProcessor.addObjectToContent` assigns ids during the initial struct-tree walk, but the subsequent `TextLineProcessor` / `ParagraphProcessor.processParagraphs` post-processing materialises new top-level objects (`SemanticParagraph`, `ImageChunk`, `PDFList`, `TableBorder`, …) that miss that hook. Those new objects are then serialised without an `id`.

Add a final per-page id-fill pass at the end of `TaggedDocumentProcessor.processDocument` that walks each page's top-level items and recurses into the recognised containers (`SemanticHeaderOrFooter`, `PDFList`, `ListItem`, `TableBorder`), assigning an id only when one is missing. The helper is intentionally local to the tagged path; promoting it into `DocumentProcessor.setIDs` would change the non-tagged path's id ordering, which `remapMetadataToContents` and `CaptionProcessor` depend on.

## Evidence
Ran the CLI on `samples/pdf/pdfua-1-reference-suite-1-1/PDFUA-Ref-2-01_Magazine-danish.pdf` with and without `--use-struct-tree` and audited every JSON node by `type`.

| `--use-struct-tree`, before fix | id-mandatory types may be missing ids | image 67/98, paragraph 291/332, text chunk 0/28 — ~70 items missing |
| `--use-struct-tree`, after fix | every id-mandatory type 100% | image 98/98, paragraph 332/332, heading 70/70, list 6/6, list item 30/30, footer 28/28, text chunk 28/28 |
| default mode, before vs after | unchanged | per-type counts byte-identical; top-level item type sequence diff = 0 |
| full core test suite | green | 645/645 (was 643 + 2 new regression tests) |

New regression test `PdfDlospStructTreeIdsIntegrationTest` covers both modes and fails on the pre-fix build (69 offending paths) while passing on this branch.

## Summary by CodeRabbit

* **New Features**
  * PDF documents now receive complete structure identification with proper ID assignment for all recognized content types when using struct tree mode.

* **Tests**
  * Added integration tests to verify complete ID assignment for recognized content items and validate struct tree mode functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opendataloader-project/opendataloader-pdf/pull/505)

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/TaggedDocumentProcessor.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/TaggedDocumentProcessor.java
@@ -60,7 +60,70 @@ public class TaggedDocumentProcessor {
             List<IObject> pageContents = TextLineProcessor.processTextLines(contents.get(pageNumber));
             contents.set(pageNumber, ParagraphProcessor.processParagraphs(pageContents));
         }
+        // PDFDLOSP-7: ParagraphProcessor (and earlier helpers) materialise new
+        // top-level objects (paragraphs, images, list/table containers) that
+        // did not exist when addObjectToContent assigned ids during struct-tree
+        // traversal. Walk the final per-page contents and fill in any missing
+        // ids. Items that already have one are left untouched so the order
+        // produced by the struct-tree pass is preserved.
+        for (int pageNumber = 0; pageNumber < totalPages; pageNumber++) {
+            if (shouldProcessPage(pageNumber)) {
+                fillMissingIds(contents.get(pageNumber));
+            }
+        }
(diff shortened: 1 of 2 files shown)


bundolee:
Closing for follow-up — minor review polish to apply (commit scope rename + dead-code comment) before re-opening.

PR #521, opened 2026-05-20
feat(python): add optional timeout kwarg to convert() / run_jar()

tensorboy:
Adds an optional `timeout: Optional[float] = None` kwarg to both the
low-level `run_jar()` and the auto-generated `convert()`.

When set, the JVM is bounded to that many wall-clock seconds; on
expiry it gets SIGKILL and `subprocess.TimeoutExpired` propagates to
the caller — same exception shape `subprocess.run` already raises,
so existing `except` handlers stay uniform across quiet and streaming
modes.

## Why

A long-running service that imports `opendataloader-pdf` has no upper
bound on a single JAR invocation today. `subprocess.run()` inside
`run_jar()` is called without `timeout=`, so a single pathological
PDF that sends the JVM into a tight loop (or just runs slow on a
small host) blocks the whole calling process until somebody kills
the Java PID by hand.

I hit this 3 times in production this past week on an OCR
microservice that runs `opendataloader_pdf.convert()` inside an
asyncio FastAPI endpoint with `uvicorn --workers 1`. One slow/hung
PDF blocked **every** subsequent request indefinitely — worst
observed hang was 1h18m before manual intervention. My workaround
was to bypass `convert()` and call `subprocess.run` on the JAR
myself with `timeout=` — works fine but loses the typed
`convert()` API and means every consumer reinvents this wheel.

## Behaviour

```python
opendataloader_pdf.convert("in.pdf", timeout=90)                      # quiet (default)
opendataloader_pdf.convert("in.pdf", quiet=False, timeout=90)         # streaming
```

The kwarg threads through both subprocess paths in `run_jar()`:

- **quiet** forwards `timeout=` straight to `subprocess.run`, which
  already supports it natively. SIGKILL + `TimeoutExpired` is the
  stdlib behaviour, no custom code needed.
- **streaming** cannot use `subprocess.run`’s timeout (it iterates
  `process.stdout` line-by-line as the JVM emits), so it tracks a
  monotonic deadline, calls `process.kill()` on expiry, and raises
  the same `subprocess.TimeoutExpired` shape. Output gathered before
  the kill is preserved in `exc.output`.

`timeout` is intentionally a Python-runtime concept and lives
outside `options.json` — it doesn’t map to a Java CLI flag (the
JAR itself has no self-imposed deadline). The generator
(`scripts/generate-options.mjs`) emits it as a separate kwarg at
the end of `convert()`’s signature with a docstring covering the
SIGKILL + `TimeoutExpired` semantics.

Default `timeout=None` preserves every existing caller’s behaviour
verbatim — the kwarg is purely opt-in.

## Tests

Three new cases in `tests/test_runner.py`:

1. quiet mode forwards `timeout=` to `subprocess.run`
2. streaming mode kills the process on expiry and raises
   `TimeoutExpired` with the same shape as the quiet path
3. `timeout=None` preserves legacy behaviour (kwarg present but not
   treated specially — flushes through as `None`)

All 5 runner tests pass locally.

## Files

- `scripts/generate-options.mjs` — generator emits the new kwarg
  (regenerated `convert_generated.py` is in this PR for review even
  though it is auto-generated)
- `python/opendataloader-pdf/src/opendataloader_pdf/runner.py`
- `python/opendataloader-pdf/src/opendataloader_pdf/convert_generated.py` (regenerated)
- `python/opendataloader-pdf/tests/test_runner.py`

Happy to follow up with the same kwarg on the Node binding if
youd like — left it out of this PR to keep the diff focused.

## Summary by CodeRabbit

* **New Features**
  * Added optional timeout parameter to PDF conversion operations, enabling users to set wall-clock limits (in seconds) for processing with automatic subprocess termination on timeout and enhanced error handling

* **Tests**
  * Added comprehensive unit tests covering timeout behavior in both quiet and streaming modes with proper error propagation

[Review Change Stack](https://app.coderabbit.ai/change-stack/opendataloader-project/opendataloader-pdf/pull/521?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/python/opendataloader-pdf/src/opendataloader_pdf/runner.py
+++ b/python/opendataloader-pdf/src/opendataloader_pdf/runner.py
@@ -4,14 +4,40 @@ Low-level JAR runner for opendataloader-pdf.
 import subprocess
 import sys
 import importlib.resources as resources
-from typing import List
+from typing import List, Optional
 
 # The consistent name of the JAR file bundled with the package
 _JAR_NAME = "opendataloader-pdf-cli.jar"
 
 
-def run_jar(args: List[str], quiet: bool = False) -> str:
-    """Run the opendataloader-pdf JAR with the given arguments."""
+def run_jar(
+    args: List[str],
+    quiet: bool = False,
+    timeout: Optional[float] = None,
+) -> str:
+    """Run the opendataloader-pdf JAR with the given arguments.
+
+    Args:
+        args: Arguments forwarded to the bundled JAR.
+        quiet: Suppress live console streaming and capture stdout instead.
+        timeout: Optional wall-clock timeout in seconds. If the JVM hasn't
+            finished by then it is sent SIGKILL and ``subprocess.TimeoutExpired``
+            is raised. ``None`` (default) preserves the original
+            no-timeout behaviour. Useful when invoking the JAR from a
+            long-running service where a single malformed/pathological PDF
+            must not be able to stall the caller indefinitely.
+
+    Returns:
+        Captured stdout from the JAR. In streaming mode this is the same
+        text that was tee-d to ``sys.stdout`` during the run.
+
+    Raises:
+        FileNotFoundError: ``java`` is not on PATH.
+        subprocess.CalledProcessError: JAR exited non-zero.
+        subprocess.TimeoutExpired: ``timeout`` was set and elapsed. The
+            JVM has been killed by the time this is raised; partial
+            output is in ``exc.stdout``.
+    """
     try:
         # Access the embedded JAR inside the package
         jar_ref = resources.files("opendataloader_pdf").joinpath("jar", _JAR_NAME)
@@ -30,7 +56,9 @@ def run_jar(args: List[str], quiet: bool = False) -> str:
             ]
 
             if quiet:
-                # Quiet mode → capture all output
+                # Quiet mode → capture all output. ``subprocess.run`` natively
+                # supports ``timeout``; on expiry it sends SIGKILL to the
+                # child and raises ``TimeoutExpired`` — no orphan JVM.
                 result = subprocess.run(
                     command,
                     capture_output=True,
@@ -38,10 +66,18 @@ def run_jar(args: List[str], quiet: bool = False) -> str:
                     check=True,
                     encoding="utf-8",
                     errors="replace",
+                    timeout=timeout,
                 )
                 return result.stdout
 
-            # Streaming mode → live output
+            # Streaming mode → live output. We can't rely on subprocess.run's
+            # timeout here because we're tee-ing lines as they arrive. Track
+            # the deadline manually and SIGKILL on expiry, then raise the
+            # same ``TimeoutExpired`` shape so callers can ``except`` uniformly
+            # across quiet/streaming modes.
+            import time as _time
+            deadline = (_time.monotonic() + timeout) if timeout is not None else None
+
             with subprocess.Popen(
                 command,
                 stdout=subprocess.PIPE,
@@ -51,15 +87,36 @@ def run_jar(args: List[str], quiet: bool = False) -> str:
                 errors="replace",
             ) as process:
                 output_lines: List[str] = []
-                for line in process.stdout:
-                    if hasattr(sys.stdout, "buffer"):
-                        sys.stdout.buffer.write(line.encode("utf-8", errors="replace"))
-                        sys.stdout.buffer.flush()
-                    else:
-                        sys.stdout.write(line)
-                    output_lines.append(line)
+                try:
+                    for line in process.stdout:
+                        if hasattr(sys.stdout, "buffer"):
+                            sys.stdout.buffer.write(
+                                line.encode("utf-8", errors="replace")
+                            )
+                            sys.stdout.buffer.flush()
+                        else:
+                            sys.stdout.write(line)
+                        output_lines.append(line)
+                        if deadline is not None and _time.monotonic() > deadline:
+                            process.kill()
+                            captured = "".join(output_lines)
+                            raise subprocess.TimeoutExpired(
+                                command, timeout, output=captured
+                            )
 
-                return_code = process.wait()
+                    # ``timeout`` here covers the case where the process
+                    # has finished streaming but is still in the wait()
+                    # finalization (rare but possible).
+                    if deadline is not None:
+                        remaining = max(0.0, deadline - _time.monotonic())
+                    else:
+                        remaining = None
+                    return_code = process.wait(timeout=remaining)
+                except subprocess.TimeoutExpired:
+                    process.kill()
+                    process.wait()
+                    raise
+
                 captured_output = "".join(output_lines)
 
                 if return_code:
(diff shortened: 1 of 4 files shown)


python/opendataloader-pdf/tests/test_runner.py
```diff
@@ -65,6 +65,77 @@ def test_streaming_failure_does_not_duplicate_output(monkeypatch, capsys, patche
...
+    fake_process.__enter__ = lambda self: self
+    fake_process.__exit__ = lambda self, *_a: False
+
+    monkeypatch.setattr(runner.subprocess, "Popen", lambda *_a, **_kw: fake_process)
+
+    # Inject a fake time source that advances faster than the loop reads
+    # lines, so the deadline expires on the first iteration.
+    import itertools
+    t = itertools.count(start=0, step=10)
+    monkeypatch.setattr(runner, "_time" if False else "subprocess",
```
hnc-jglee, line 107: Small correctness nit: the ternary `"_time" if False else "subprocess"` always evaluates to `"subprocess"`, so this line ends up reassigning `runner.subprocess` to itself rather than patching `time.monotonic` as it looks like it intends to.

The practical effect: the streaming deadline branch in `runner.py` is never actually triggered by this test — the fake `itertools.count` clock is wired to `_real_time.monotonic` on the next line but `runner.py` reads `_time.monotonic()` from its own local `import time as _time`, which still uses the real clock. The test passes without exercising the new logic, so a future regression in the deadline path would go undetected.

A direct patch on `time.monotonic` works as intended:

```python
import itertools, time as _time_module
t = itertools.count(start=0, step=10)
monkeypatch.setattr(_time_module, "monotonic", lambda: next(t))
```

And two extra assertions would lock the behavior in:

```python
fake_process.kill.assert_called()
assert "chunk" in excinfo.value.output  # partial output preserved on TimeoutExpired
```

python/opendataloader-pdf/src/opendataloader_pdf/runner.py
```diff
@@ -51,15 +87,36 @@ def run_jar(args: List[str], quiet: bool = False) -> str:
...
+                        if hasattr(sys.stdout, "buffer"):
+                            sys.stdout.buffer.write(
+                                line.encode("utf-8", errors="replace")
+                            )
+                            sys.stdout.buffer.flush()
+                        else:
+                            sys.stdout.write(line)
+                        output_lines.append(line)
+                        if deadline is not None and _time.monotonic() > deadline:
+                            process.kill()
```
hnc-jglee, line 101: Small reliability concern: raising `TimeoutExpired` from inside the `with subprocess.Popen(...)` block does run `Popen.__exit__`, but since the process has just been killed, the underlying `communicate()`-based cleanup doesn't fully close the pipes, and `process.stdout`'s file descriptor leaks. In one-off use this is invisible, but given the PR's motivation (single-worker FastAPI under sustained load), repeated timeouts can gradually fill the FD table until subsequent JVM spawns fail with `EMFILE` / "too many open files".

Explicitly closing stdout and reaping the process before raising keeps things tidy:

```python
if deadline is not None and _time.monotonic() > deadline:
    process.kill()
    try:
        process.stdout.close()
    except Exception:
        pass
    process.wait()  # reap so we don't leave a zombie
    captured = "".join(output_lines)
    raise subprocess.TimeoutExpired(command, timeout, output=captured)
```

python/opendataloader-pdf/src/opendataloader_pdf/runner.py
```diff
@@ -51,15 +87,36 @@ def run_jar(args: List[str], quiet: bool = False) -> str:
...
+                    # ``timeout`` here covers the case where the process
+                    # has finished streaming but is still in the wait()
+                    # finalization (rare but possible).
+                    if deadline is not None:
+                        remaining = max(0.0, deadline - _time.monotonic())
                     else:
-                        sys.stdout.write(line)
-                    output_lines.append(line)
+                        remaining = None
+                    return_code = process.wait(timeout=remaining)
```
hnc-jglee, line 114: Small correctness nit: `process.wait(timeout=remaining)` here can itself raise `TimeoutExpired` (e.g. when `remaining ≈ 0` and the JVM is still finalizing). The surrounding `except subprocess.TimeoutExpired` block below does handle it, but explicit handling at the call site makes the cleanup contract clearer and keeps caller block time predictable:

```python
try:
    return_code = process.wait(timeout=remaining)
except subprocess.TimeoutExpired:
    process.kill()
    process.wait()
    captured = "".join(output_lines)
    raise subprocess.TimeoutExpired(command, timeout, output=captured)
```

This also ensures partial output is preserved on the wait-timeout path, the same way the streaming-loop timeout path does.

python/opendataloader-pdf/src/opendataloader_pdf/runner.py
```diff
@@ -4,14 +4,40 @@
...
 # The consistent name of the JAR file bundled with the package
 _JAR_NAME = "opendataloader-pdf-cli.jar"
 
 
-def run_jar(args: List[str], quiet: bool = False) -> str:
-    """Run the opendataloader-pdf JAR with the given arguments."""
+def run_jar(
+    args: List[str],
+    quiet: bool = False,
+    timeout: Optional[float] = None,
```
hnc-jglee, line 16: Small input-validation suggestion: `timeout=-1` (or any negative value) currently makes the deadline already-expired, which silently means "fire immediately" on the first streaming line. That's surprising for a caller who just typo'd the value.

A tiny guard plus a docstring note would prevent the silent failure mode:

```python
if timeout is not None and timeout <= 0:
    raise ValueError(f"timeout must be > 0 seconds, got {timeout!r}")
```

(If you intentionally want `timeout == 0` to mean "fire immediately," rejecting just `< 0` and documenting that in the docstring works too — either policy is fine, just nice to have it explicit.)

hnc-jglee (commented):
Thanks for this contribution! 🙏 The motivation (a single pathological PDF hanging an asyncio/FastAPI service for 1h18m) really resonates, and delegating the quiet path to `subprocess.run(timeout=...)` is a clean choice. Keeping `timeout` out of `options.json` as a Python-runtime parameter is also the right separation of concerns.

We did a multi-perspective review (code quality, tests, architecture, reliability) and have **four small must-fix items inline**, all scoped to the streaming path — the quiet path looks production-ready as-is.

**Inline summary:**
1. `tests/test_runner.py:107` — the `"_time" if False else "subprocess"` ternary always resolves to `"subprocess"`, so `time.monotonic` is never patched and the streaming deadline branch is never actually exercised by the test (it passes without validating the core new logic).
2. `runner.py:101` — raising `TimeoutExpired` inside the `with subprocess.Popen(...)` block skips proper cleanup of `process.stdout`; the pipe FD leaks. Under repeated timeouts in a single-worker FastAPI, this can hit `EMFILE` on subsequent JVM spawns.
3. `runner.py:114` — the post-loop `process.wait(timeout=remaining)` can itself raise `TimeoutExpired`; wrapping it in its own `try/except` keeps caller block time predictable.
4. `runner.py:16` — there's no guard against `timeout <= 0`, so `timeout=-1` silently means "fire immediately". A small `ValueError` plus a docstring note would prevent surprises.

**Suggested follow-ups (not blocking this PR):**
- Process-group isolation (`start_new_session=True` + `os.killpg`) so SIGKILL reaches any JVM grandchildren.
- A docstring note that `process.kill()` + `process.wait()` are synchronous and can briefly block an asyncio event loop.
- Truncating `exc.output` to ~1 MB to bound memory on repeated timeouts.
- Node binding parity for the same `timeout` kwarg (which you already mentioned in the PR description).
- A declarative `RUNTIME_ONLY_KWARGS` map in `generate-options.mjs` so future kwargs (`retry`, `jvm_heap`, …) don't accumulate inline special cases.
- A one-line note in CLAUDE.md that runtime-only kwargs live in `scripts/generate-options.mjs`, not `options.json` (so they bypass the `npm run sync` gate).

Happy to split any of these into separate issues if that helps keep this PR focused. Thanks again! 🙏

PR #524, opened 2026-05-20
Update HTML output with layout attributes

LonelyMidoriya:
**Checklist:**

## Summary by CodeRabbit

* **New Features**
  * Enhanced HTML output with support for styled text formatting, including italic, strikethrough, custom font colors, and font weights.
  * Added text alignment support in HTML generation.

* **Chores**
  * Updated VeraPDF dependencies to improve PDF handling capabilities.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opendataloader-project/opendataloader-pdf/pull/524?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/html/HtmlGenerator.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/html/HtmlGenerator.java
@@ -36,8 +36,11 @@ import org.verapdf.wcag.algorithms.entities.lists.PDFList;
 import org.verapdf.wcag.algorithms.entities.tables.tableBorders.TableBorder;
 import org.verapdf.wcag.algorithms.entities.tables.tableBorders.TableBorderCell;
 import org.verapdf.wcag.algorithms.entities.tables.tableBorders.TableBorderRow;
+import org.verapdf.wcag.algorithms.semanticalgorithms.consumers.ContrastRatioConsumer;
 import org.verapdf.wcag.algorithms.semanticalgorithms.containers.StaticContainers;
+import org.verapdf.wcag.algorithms.semanticalgorithms.utils.NodeUtils;
 
+import java.awt.*;
 import java.io.Closeable;
 import java.io.File;
 import java.io.FileWriter;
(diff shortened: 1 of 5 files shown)


java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/html/HtmlGenerator.java
```diff
@@ -494,16 +496,38 @@ protected String escapeHtmlAttribute(String value) {
...
 
+    private static String getTextStyle(TextChunk chunk) {
+        StringBuilder style = new StringBuilder();
+        if (chunk.getIsStrikethroughText()) {
+            style.append(HtmlSyntax.HTML_STRIKETHROUGH_STYLE_PROPERTY);
+        }
+        if (chunk.isItalic()) {
+            style.append(HtmlSyntax.HTML_ITALIC_STYLE_PROPERTY);
+        }
+        if (!Color.BLACK.equals(chunk.getTextColor())) {
```
hnc-jglee: Small correctness nit: `Color.equals()` compares the full RGBA object, so a `Color` extracted via `getTextColor()` that is *visually* black but constructed as a different instance (or carries a different alpha) won't match `Color.BLACK` here, and we'll end up emitting a `<span>` for plain black text.

Would you consider comparing the packed RGB int instead? For example:

```java
Color textColor = chunk.getTextColor();
if (textColor != null && (textColor.getRGB() & 0x00FFFFFF) != (Color.BLACK.getRGB() & 0x00FFFFFF)) {
    ...
}
```

Also — just to double-check — this branch reads `chunk.getTextColor()` for the comparison but then uses `chunk.getFontColor()` on the next line to actually build the CSS value. Is that intentional, or should they be the same source? A short comment clarifying which one is authoritative would be very helpful for future readers.

java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/html/HtmlGenerator.java
```diff
@@ -494,16 +496,38 @@ protected String escapeHtmlAttribute(String value) {
...
+            Color color = ContrastRatioConsumer.getTextColorFromComponentArray(chunk.getFontColor());
```
hnc-jglee: Two small defensive concerns here:

1. **Null safety** — if `chunk.getFontColor()` is null or the underlying component array is unexpected, `getTextColorFromComponentArray(...)` may return `null`, and the very next line (`color.getRed()`) would NPE. A guard with a fallback to black would make this robust:
   ```java
   Color color = ContrastRatioConsumer.getTextColorFromComponentArray(chunk.getFontColor());
   if (color == null) {
       color = Color.BLACK;
   }
   ```

2. **RGB bounds** — depending on what the WCAG layer returns, the per-channel values might fall outside `0..255` in edge cases, which would produce invalid CSS like `color: rgb(300, -5, 256);`. Clamping is a one-liner:
   ```java
   int r = Math.max(0, Math.min(255, color.getRed()));
   int g = Math.max(0, Math.min(255, color.getGreen()));
   int b = Math.max(0, Math.min(255, color.getBlue()));
   ```

One related question: since `HtmlGenerator` runs under `ForkJoinPool(availableProcessors)` for per-page work, do you happen to know whether `ContrastRatioConsumer` is safe to call from worker threads here? CLAUDE.md notes a similar consumer can't be parallelized safely, so a quick JavaDoc note on this method's threading assumptions would help.

java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/html/HtmlGenerator.java
```diff
@@ -494,16 +496,38 @@ protected String escapeHtmlAttribute(String value) {
...
+            style.append(String.format(HtmlSyntax.HTML_FONT_COLOR_PROPERTY,
+                color.getRed(), color.getGreen(), color.getBlue()));
+        }
+        int fontWeight = chunk.getRoundedFontWeight();
+        if (fontWeight != 400) {
```
hnc-jglee, line 524: Tiny robustness suggestion: CSS only treats `font-weight` values in the `100..900` range as valid. If `getRoundedFontWeight()` ever returns `0`, `-1`, or something outside that range (e.g. for a chunk where weight wasn't resolved), the current code will still emit `font-weight: -1;`, which most browsers will silently drop.

A small range guard would keep the output spec-compliant:

```java
int fontWeight = chunk.getRoundedFontWeight();
if (fontWeight >= 100 && fontWeight <= 900 && fontWeight != 400) {
    style.append(String.format(HtmlSyntax.HTML_FONT_WEIGHT_PROPERTY, fontWeight));
}
```

java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/utils/GeneratorUtils.java
```diff
@@ -8,6 +8,7 @@
 import org.verapdf.wcag.algorithms.entities.content.TextLine;
 import org.verapdf.wcag.algorithms.semanticalgorithms.utils.TextChunkUtils;
 
+import java.awt.*;
```
hnc-jglee: Nit: this `import java.awt.*;` doesn't appear to be used anywhere in `GeneratorUtils.java` — could it be removed? (Or if it's needed by an upcoming change in this PR, no worries.)

hnc-jglee (commented):
Thanks for working on this — consolidating strikethrough/italic/color/weight into a single `getTextStyle()` is a nice cleanup, and the conditional `<span>` wrapping is a good touch.

We did a multi-perspective review (code quality, architecture, test coverage, security) and would like to suggest a small set of must-fix items inline before merge. All of them are low-effort defensive tweaks plus adding unit tests for the new method. Nothing fundamental — the design direction looks good.

**Summary of inline comments:**
1. `HtmlGenerator.java:519` — `Color.equals()` compares the full RGBA object; suggest comparing RGB ints so a different `Color` instance of "black" still skips the span.
2. `HtmlGenerator.java:520` — `getTextColorFromComponentArray()` may return `null` / out-of-range values; suggest a null guard + clamping to `[0, 255]`. Worth a note on thread-safety too since this runs inside `ForkJoinPool`.
3. `HtmlGenerator.java:524-526` — `font-weight` is emitted unbounded; CSS only accepts `100..900`, so a range guard would prevent invalid output.
4. `GeneratorUtils.java:11` — the new `import java.awt.*` doesn't seem to be used in this file.

**One follow-up suggestion (not blocking):** since this is the first piece of inline styling in the HTML generator, adding a small `HtmlGeneratorTest` covering the 2^4 style combinations + a few null/edge cases would lock in the behavior and help future contributors. Happy to discuss splitting that into a follow-up issue if you'd prefer to keep this PR focused.

Thanks again for the contribution! 🙏

LonelyMidoriya:
I fixed possible NPE caused by color and implemented your suggestion about Alpha in Color. We have check for RGB and font-weight bounds in verapdf. Also I've added HtmlGeneratorTest covering the 2^4 style combinations + a few null/edge cases as you suggested. Thank you for reviewing this PR and helping with fixes.

MaximPlusov (approved):

PR #526, opened 2026-05-20
feat(hybrid): expose raw response + reference environment + client wall-clock for downstream tooling

bundolee:
## Summary

- Expose three new static volatiles + getters on `HybridDocumentProcessor`: `lastHybridRawJson`, `lastHybridHealth`, `lastHybridClientMs`
- Add `HybridClient.fetchHealth()` default method; `HancomAIClient` overrides it (GET `/health`)
- Copy top-level `confidence` from TSR + IMAGE_CAPTIONING through merged JSON so downstream can surface per-table / per-caption self-scores
- Rename DLA label constants to the authoritative hancom-ai names, add codes 5 (Date) and 16 (Number)

Required by opendataloader-pdfua [`feat/evidence-report-series`](https://github.com/opendataloader-project/opendataloader-pdfua/pull/55) — the new evidence-report needs all three hooks to persist per-module raw responses, reference hardware, and honest SLA-grade timings.

## Why client wall-clock alongside server-reported timings

`HybridResponse.getTimings()` reports on-GPU work the server self-reports. Downstream tooling reasoning about SLA / throughput needs the figure the *user* actually waited on — network RTT, JSON (de)serialization, queueing. `lastHybridClientMs` accumulates around each `client.convert(...)` in a `try-finally`, so failures (timeouts, 5xx) are accounted too. Real-server and mock-fixture runs are now directly comparable on the same client-side scale, which the server-reported `total_ms` isn't (fixture-replay servers return canned timings).

## Test plan

- [x] `mvn install -DskipTests` from `java/` — BUILD SUCCESS
- [x] Single-threaded-per-process contract preserved (same volatile-static pattern as the existing `lastHybridTimings`)
- [x] `fetchHealth()` default returns null; other `HybridClient` implementations (Docling, Hancom) inherit unchanged
- [x] No production / test code pins the old DLA label constants by string
- [x] Two rounds of independent senior review (✅ ready to merge)
- [ ] CI pipeline (will run on PR push)

## Notes

- Single squashed commit. Rebased onto current `main`.
- Multi-chunk documents: `lastHybridClientMs` accumulates across chunks; `lastHybridRawJson` and `lastHybridTimings` keep last-chunk-wins (existing contract).

## Summary by CodeRabbit

* **New Features**
  * Optional health-check support for hybrid AI backends (best-effort, non-fatal)
  * Image captioning now returns optional model confidence per figure
  * Table extraction includes optional confidence per table entry
  * Telemetry extended to record client-side processing time, last backend response, and last backend health snapshot

* **Tests**
  * Added health-check tests covering success, missing health, non-JSON responses, and unreachable server cases

[Review Change Stack](https://app.coderabbit.ai/change-stack/opendataloader-project/opendataloader-pdf/pull/526?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/hybrid/HancomAIClient.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/hybrid/HancomAIClient.java
@@ -67,6 +67,7 @@ public class HancomAIClient implements HybridClient {
     private static final String SDK_ENDPOINT = "/hocr/sdk";
     private static final String PDF2IMG_ENDPOINT = "/support/pdf2img";
     private static final String PING_ENDPOINT = "/ping";
+    private static final String HEALTH_ENDPOINT = "/health";
     private static final int HEALTH_CHECK_TIMEOUT_MS = 3000;
     private static final String DEFAULT_FILENAME = "document.pdf";
     private static final MediaType MEDIA_TYPE_PDF = MediaType.parse("application/pdf");
@@ -135,6 +136,28 @@ public class HancomAIClient implements HybridClient {
         this.baseUrl = baseUrl;
         this.httpClient = httpClient;
         this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public JsonNode fetchHealth() {
+        OkHttpClient healthClient = httpClient.newBuilder()
+            .connectTimeout(HEALTH_CHECK_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+            .readTimeout(HEALTH_CHECK_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+            .build();
+        Request request = new Request.Builder()
+            .url(baseUrl + HEALTH_ENDPOINT)
+            .get()
+            .build();
+        try (Response response = healthClient.newCall(request).execute()) {
+            if (!response.isSuccessful()) return null;
+            ResponseBody body = response.body();
+            if (body == null) return null;
+            return objectMapper.readTree(body.string());
+        } catch (IOException e) {
+            LOGGER.log(Level.FINE,
+                "Hancom AI /health unavailable: {0}", e.getMessage());
+            return null;
+        }
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


bundolee:
Thanks for the review. All four findings addressed in `5b0cf886` (force-pushed):

1. **`isNumber()` guard on confidence reads** — applied at both sites (TSR per-table `entry.put("confidence", …)` at `HancomAIClient.java:505` and the captioning return at `:716-718`). Confidence is now only carried through when the node is actually numeric; non-numeric/null silently omit the field instead of being coerced to 0.0.

2. **Explicit `LABEL_DATE` / `LABEL_NUMBER` cases** — added next to `LABEL_PARA_TEXT` / `LABEL_OTHER_TEXT` in `HancomAISchemaTransformer.java:412-419`. Both labels now flow through `createParagraph` with an explicit case, so they aren't dependent on the default branch and don't leave their constants "unused".

3. **`Throwable` → `Exception` on health update** — `HybridDocumentProcessor.java:580-587`. JVM-fatal errors (OOM, StackOverflow) propagate; downstream-fixable IO/parse failures still degrade gracefully to `lastHybridHealth = null`. Added a FINE log with the caught exception for diagnosis.

4. **`System.nanoTime()` for the client wall-clock** — `HybridDocumentProcessor.java:620-633`. Monotonic source, converted to milliseconds via `TimeUnit.NANOSECONDS.toMillis` before adding to `lastHybridClientMs`. Variable renamed `convertStartNs` to make the unit unambiguous.

Plus, separately, four new tests in `HealthCheckTest` covering `HancomAIClient.fetchHealth()` (success / 404 / non-JSON body / server down) — all four assert null-not-throw semantics, which is the contract the processor's `catch (Exception)` relies on. Local `mvn test`: 10/10 in `HealthCheckTest`, full suite green.

bundolee:
Both nitpicks addressed in `7c4620a3`:

- **`asDouble()` → `doubleValue()`** at `HancomAIClient.java:507` and `:721` after the `isNumber()` guard. The guard makes the 0.0 fallback unreachable in practice, but `doubleValue()` makes the intent explicit. Short comment added at the TSR site explaining the choice.
- **`LABEL_DATE` / `LABEL_NUMBER` comment refined** at `HancomAISchemaTransformer.java:414-422`. Replaced "Page 42" with "€42.00" (PAGE_NUMBER vs NUMBER confusion was real), and reworded "disappearing into the default branch" → "keeps the routing intent visible and prevents future silent misrouting" since the default branch also creates paragraphs today.

Local `mvn test` still green.

PR #527, opened 2026-05-20
fix(processors): surface clean error for corrupted or truncated PDFs

hyunhee-jo:
## Objective

A PDF whose `%PDF-` header is present but whose body is corrupt or truncated (download interrupted near the end, missing trailing xref) currently surfaces a 17-line veraPDF `IOException` stack trace on stdout. The user has no way to tell whether their file is broken or whether opendataloader-pdf is.

## Approach

Wrap only the `new PDDocument(pdfName)` call in `DocumentProcessor.preprocessing` with a try-catch on `IOException` and rethrow as `InvalidPdfFileException` (introduced by the prior wrong-content fix) with a distinct message ending in `(corrupted or truncated content)`.

`InvalidPasswordException` is re-thrown as-is so the existing password-handling branch in `CLIMain` still wins.

Other failure modes (`parseStructureTreeRoot`, `parseChunks`, hybrid path) remain on the existing SEVERE path — only the header-present-body-broken case is reclassified.

Because `InvalidPdfFileException` is already routed by `CLIMain.processFile`, no CLI changes are needed: CLI-argument input prints `Error: ...` on stdout and exits 1; a corrupt file inside a directory logs WARNING and continues (preserves batch-folder semantics). The original veraPDF exception is preserved as `getCause()` so Java API consumers can still inspect the root cause for diagnostics.

## Evidence

`mvn -pl opendataloader-pdf-cli -am test` runs `DocumentProcessorMagicNumberTest` 4/4 pass and `CLIMainTest` 23/23 pass.

Manual end-to-end against the built jar:

| Real PDF with last 2000 bytes cut off | friendly stdout, exit 1, no stack trace | `Error: 'real_truncated.pdf' is not a valid PDF file (corrupted or truncated content).` exit 1 |
| JPEG renamed to `.pdf` | distinct missing-header wording | `Error: 'fake.pdf' is not a valid PDF file (missing %PDF- header).` exit 1 |
| Intact PDF | exit 0, JSON written | exit 0, JSON generated |
| Directory with 1 truncated + 1 intact | intact processed, truncated skipped with WARNING, exit 0 | WARNING `'truncated.pdf' ... Skipping.`, `good.json` written, exit 0 |

`npm run sync` not required — no CLI option definitions changed. Python/Node wrappers pass Java stdout and exit code through unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer, user-facing error messages distinguishing missing PDF header vs. “corrupted or truncated content”.
  * Password-protected PDFs rethrown appropriately; corrupt-but-headered files report friendly errors without leaking internal stack traces. Directory-contained corrupt files are now skipped with a warning and successful exit.

* **Documentation**
  * Updated exception docs to describe the two PDF failure modes and preserved underlying causes.

* **Tests**
  * Added regression tests covering corrupt-but-headered PDFs, truncated files, and directory-skip behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opendataloader-project/opendataloader-pdf/pull/527?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/java/opendataloader-pdf-cli/src/test/java/org/opendataloader/pdf/cli/CLIMainTest.java
+++ b/java/opendataloader-pdf-cli/src/test/java/org/opendataloader/pdf/cli/CLIMainTest.java
@@ -74,14 +74,22 @@ class CLIMainTest {
 
     /**
      * When a directory contains a file that fails processing, run() must return
-     * non-zero, even though other files in the directory may succeed.
+     * non-zero, even though other files in the directory may succeed. The
+     * "failure" here must be a genuine processing error (unreachable hybrid
+     * backend), not a corrupt-content classification — corrupt PDFs inside a
+     * directory are silently skipped by design (see
+     * {@link #testCorruptPdfInsideDirectoryIsSilentlySkipped()}), so the test
+     * uses a parseable PDF and lets the hybrid availability check fail.
      */
     @Test
     void testDirectoryWithFailingFileReturnsNonZeroExitCode() throws IOException {
         Path dir = tempDir.resolve("docs");
         Files.createDirectory(dir);
         Path testPdf = dir.resolve("bad.pdf");
-        Files.write(testPdf, "%PDF-1.4 minimal".getBytes());
+        try (PDDocument doc = new PDDocument()) {
+            doc.addPage(new PDPage());
+            doc.save(testPdf.toFile());
+        }
 
         int exitCode = CLIMain.run(new String[]{
             "--hybrid", "docling-fast",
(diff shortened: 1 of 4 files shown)


hnc-jglee:
Nice cleanup — the friendly error message is a real UX win, and routing through `InvalidPdfFileException` keeps the existing CLI contract (`Error:` + exit 1 for arg mode, WARNING + exit 0 for directory mode) intact. Preserving the original veraPDF exception as `getCause()` is a nice touch for downstream diagnostics.

One thing I'd love your thoughts on before merge, if you don't mind:

**Potential file-handle leak when `new PDDocument(pdfName)` throws** — `DocumentProcessor.java:543`

```java
try {
    pdDocument = new PDDocument(pdfName);  // veraPDF opens streams inside the constructor
} catch (IOException cause) {
    throw new InvalidPdfFileException(..., cause);
}
```

Because veraPDF opens file streams inside the `PDDocument` constructor, if the constructor throws partway through, the local `pdDocument` reference is never bound — so the `finally`/`closePdfResources()` path has nothing to close. In single-file CLI mode this is harmless (the JVM exits), but in directory-batch mode a folder containing many corrupted PDFs could accumulate orphaned file descriptors across the run.

Would you be open to either:

1. A small explicit guard right in the catch (lowest-risk, scoped to this PR):
   ```java
   } catch (InvalidPasswordException pw) {
       throw pw;
   } catch (IOException cause) {
       if (pdDocument != null) {
           try { pdDocument.close(); } catch (IOException ignore) {}
       }
       throw new InvalidPdfFileException(..., cause);
   }
   ```
2. Or, if veraPDF's `PDDocument` is `AutoCloseable`-friendly here, wrapping the construction in try-with-resources — happy to defer that to a follow-up PR since it touches the broader preprocessing flow.

Totally fine to handle this as a follow-up issue if you'd rather keep this PR scoped to the error-message change — just wanted to flag it so it doesn't get lost. Everything else looks good to me 👍

hyunhee-jo:
Thanks for the careful read. Checked the veraPDF source (`parser:1.31.19`, the version this project depends on) before responding.

Cleanup is actually guaranteed at the veraPDF layer in **two** places.

**1) `org.verapdf.io.Reader.init()` closes the stream on any `IOException` and re-throws:**

```java
} catch (IOException e) {  // "If exception is thrown in init() someone should close document stream"
    this.parser.closeInputStream();
    throw e;
}
```

`closeInputStream()` chains to `SeekableBaseParser.closeInputStream()` → `this.source.close()` on the underlying `InternalInputStream`, which is the actual file handle. For the truncated-tail case in this PR (which fails at `findLastXRef` → `getXRefInfo`), the stream is closed before the `IOException` reaches us.

**2) `org.verapdf.pd.PDDocument(String)` itself is wrapped in a `Throwable`-scoped try/catch that calls `this.close()`:**

```java
public PDDocument(final String filename) throws IOException {
    try {
        this.catalog = new PDCatalog();
        this.document = new COSDocument(filename, this);
        checkPages();
    } catch (Throwable t) {
        this.close();
        throw t;
    }
}
```

So any path the inner Reader cleanup might miss (including `RuntimeException` or `Error`) is still caught at the constructor boundary.

I also realized the proposed inline guard wouldn't actually run, even if there were a leak — Java doesn't perform partial assignment on a throwing constructor:

```java
PDDocument pdDocument;
try {
    pdDocument = new PDDocument(pdfName);   // throws → assignment never happens
} catch (IOException cause) {
    if (pdDocument != null) { ... }          // compiler error: "might not have been initialized"
}
```

So I'd prefer to keep this PR scoped to the error-message change. If you ever spot actual fd accumulation in a batch run I'd absolutely want to revisit — but absent a reproducer, `Reader.init`'s catch plus the constructor-level `Throwable` handler already look like the safety net we'd want.

hnc-jglee (approved):

PR #554, opened 2026-06-03
fix(python): validate to-stdout+format compat and add batch_convert()

### LINKED ISSUE #451: Error handling in opendataloader.convert
### Requested feature

The idea comes from a case when you have quite a lot PDFs to convert under opendataloader.convert, there might be some errors which reuslts in an early stop. Is it possible to add a parameter to allow for a "pass" mode instead of the default "stop" mode? So if there is some errors in one PDF, just skip it and go for the next one.

### Alternatives
I know there can be compromises like to use one cell for each PDF and add the try-except structure, but I notice that the suggestion would be to cover the whole batch under one cell, that's why I am asking.

### LINKED ISSUE #492: bug(cli): --to-stdout silently produces empty output for --format json or html
### Bug

When `--to-stdout` is combined with `--format json` or `--format html`, the CLI exits successfully (exit code `0`) but writes **nothing** to stdout. There is no warning on stderr, no error message, and no validation at argument-parsing time. From the user's perspective the conversion silently fails — the only signal that something went wrong is an empty stdout.

`text` and `markdown` work as expected; the issue affects only `json` and `html`.

The relevant code path explicitly comments that these formats are not implemented, and the help text for `--to-stdout` does not document which formats are supported:

`java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/processors/DocumentProcessor.java:451-465`
```java
if (config.isOutputStdout()) {
    java.io.Writer stdoutWriter = new java.io.BufferedWriter(
        new java.io.OutputStreamWriter(System.out, java.nio.charset.StandardCharsets.UTF_8));
    if (config.isGenerateText()) {
        TextGenerator textGenerator = new TextGenerator(stdoutWriter, config);
        textGenerator.writeToText(contents);
        stdoutWriter.flush();
    } else if (config.isGenerateMarkdown()) {
        MarkdownGenerator markdownGenerator = new MarkdownGenerator(stdoutWriter, config);
        markdownGenerator.writeToMarkdown(contents);
        stdoutWriter.flush();
    }
    // JSON and HTML stdout not yet supported
    return;
}
```

`java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/api/cli/CLIOptions.java:201-202`
```java
private static final String TO_STDOUT_LONG_OPTION = "to-stdout";
private static final String TO_STDOUT_DESC = "Write output to stdout instead of file (single format only)";
```

The original commit (`7aaae2e`, *"perf: add --to-stdout option and refine text-only fast path"*) describes the feature as supporting "text and markdown formats", so JSON/HTML appears to be a known gap rather than a regression — but there is no user-facing signal that they are unsupported.

Related: #287 (closed) addressed a similar *silent-failure-with-exit-0* pattern for hybrid-backend errors. This issue is a different code path (`DocumentProcessor.generateOutputs` rather than `HybridDocumentProcessor`) but the user-visible symptom — silent empty output, exit 0, no diagnostic on stderr — is the same class of UX problem.

### Steps to reproduce

```bash
# Build the shaded CLI jar
./scripts/build-java.sh
JAR=java/opendataloader-pdf-cli/target/opendataloader-pdf-cli-*.jar

# Affected: --to-stdout + json (silent empty output, exit 0)
java -jar $JAR --to-stdout -f json samples/pdf/lorem.pdf > out.json 2>/dev/null
echo "exit=$?  size=$(wc -c < out.json)"
# → exit=0  size=0          ❌ no warning, no error

# Affected: --to-stdout + html (same behavior)
java -jar $JAR --to-stdout -f html samples/pdf/lorem.pdf > out.html 2>/dev/null
echo "exit=$?  size=$(wc -c < out.html)"
# → exit=0  size=0          ❌

# Working: text and markdown
java -jar $JAR --to-stdout -f markdown samples/pdf/lorem.pdf 2>/dev/null | wc -c
# → 462                     ✅
java -jar $JAR --to-stdout -f text samples/pdf/lorem.pdf 2>/dev/null | wc -c
# → 459                     ✅
```

Reproduced on both Linux (WSL) and Windows PowerShell with OpenJDK 21.

### Expected behavior

The user should not be left guessing whether the conversion succeeded. Any of the following would be acceptable, in increasing order of effort:

1. **Reject at argument-parse time** (preferred minimal fix). In `CLIOptions.applyAllOptionsTo` (or a dedicated validator), throw `IllegalArgumentException` when `--to-stdout` is combined with `--format json`/`html`/`pdf`/`tagged-pdf`:
   ```java
   if (config.isOutputStdout() && (config.isGenerateJSON() || config.isGenerateHtml()
           || config.isGeneratePDF() || config.isGenerateTaggedPDF())) {
       throw new IllegalArgumentException(
           "--to-stdout currently supports only --format text or markdown");
   }
   ```
   This produces a clear error and a non-zero exit code.

2. **Update the help text.** Change `TO_STDOUT_DESC` from `"Write output to stdout instead of file (single format only)"` to `"Write output to stdout instead of file. Supported formats: text, markdown."` so the constraint is discoverable without trial and error.

3. **Implement JSON/HTML stdout support** (full fix). `JsonWriter` and `HtmlGenerator` would need overloads that accept a `Writer`, mirroring how `TextGenerator` and `MarkdownGenerator` already do, and `DocumentProcessor.generateOutputs` would route them through that path.

A minimal, defensive fix is (1) + (2); (3) can follow if the feature is desired.

### Version

`main` @ `1bb3e71` (current HEAD as of report). Reproduced against a freshly built `opendataloader-pdf-cli-0.0.0.jar`.

### Java version

```
openjdk version "21.0.10" 2026-01-20 LTS
OpenJDK Runtime Environment Microsoft-13106404 (build 21.0.10+7-LTS)
OpenJDK 64-Bit Server VM Microsoft-13106404 (build 21.0.10+7-LTS, mixed mode, sharing)
```

charlesaurav13:
## Summary

Closes #492, closes #451.

Two related improvements to the Python wrapper (`opendataloader_pdf`):

### fix(python): `--to-stdout` + unsupported format raises `ValueError` early (#492)

When `to_stdout=True` is combined with `format="json"` or `format="html"`, the JAR silently exits 0 with empty output — callers have no way to detect the failure. This fix wraps the auto-generated `convert()` with a thin validation layer (using `functools.wraps` to preserve the signature and docstring) that raises a clear `ValueError` **before** the JAR is invoked.

Validated formats for `--to-stdout`: `text`, `markdown`. All others (`json`, `html`, `pdf`, `tagged-pdf`) raise:
```
ValueError: --to-stdout does not support format(s): json. Supported formats for --to-stdout: text, markdown.
```

### feat(python): add `batch_convert()` for per-file error tolerance (#451)

```python
from opendataloader_pdf import batch_convert

results = batch_convert(
    ["report1.pdf", "corrupt.pdf", "report2.pdf"],
    errors="ignore",      # or "raise" (default — same as calling convert() in a loop)
    format="text",
    output_dir="/out",
)
# results = {"report1.pdf": None, "corrupt.pdf": <CalledProcessError>, "report2.pdf": None}
```

`errors="ignore"` catches per-file exceptions, emits a `RuntimeWarning` naming the file, and records the exception in the returned dict. `errors="raise"` (default) re-raises immediately, identical to the current behavior.

## Changes

- `src/opendataloader_pdf/wrapper.py` — validation wrapper around `_convert_impl`, new `batch_convert()`
- `src/opendataloader_pdf/__init__.py` — export `batch_convert`
- `tests/test_wrapper.py` — 16 unit tests (no Java required; JAR is mocked via `monkeypatch`)

## Test results

```
16 passed in 0.03s
```

## Release Notes

* **New Features**
  * Added batch conversion capability to process multiple files in a single operation with configurable error handling modes.
  * Enhanced validation to prevent incompatible output format and mode combinations, providing clearer error guidance.


hnc-jglee (changes requested):
Thanks for tackling this — the `errors="raise"|"ignore"` convention reads naturally, the `functools.wraps` approach to preserve the generated signature is a nice touch, and the test setup (mocking the JAR via `monkeypatch`, no Java required) is exactly right.

A few things I'd like to sort out before merge:

**1. The fix is at the wrong layer to close #492.** #492 is a JAR/CLI bug — it reproduces via `java -jar` directly, and `DocumentProcessor.java` still has `// JSON and HTML stdout not yet supported` returning empty. This PR only guards the Python path, so direct CLI users and the Node binding stay broken. Our binding architecture (the `npm run sync` pipeline) is built so cross-binding rules live once in `CLIOptions` and propagate everywhere. Could we move the validation into `CLIOptions` argument parsing (alongside the existing `--format`/`--hybrid` validators) so it fails fast with a non-zero exit for *all* surfaces? The Python guard is fine to keep as defense-in-depth on top of that.

Related: hardcoding `_STDOUT_UNSUPPORTED_FORMATS` in Python means that when the JAR eventually implements json stdout, the wrapper will wrongly reject a now-valid call with no test catching it. Sourcing the rule from the JAR avoids that drift.

**2. The validation has a hole on the most common case.** The guard is `if to_stdout and fmt:`, but the JAR's default format is `json`. So `convert(input_path="a.pdf", to_stdout=True)` (no `format`) skips the check and still silently produces empty output — the exact #492 symptom. Suggest treating an absent format as `json` before the check (`fmt or "json"`), and adding a test for it.

**3. The help text (#492's part 2) is still misleading.** `TO_STDOUT_DESC` remains `"...(single format only)"`, which now contradicts the new `ValueError`. One line in `CLIOptions` + `npm run sync` would make the CLI `--help`, `options.json`, and all bindings tell the same story.

**4. CodeRabbit's positional-arg finding is real.** The wrapper reads only `kwargs.get(...)`, but `format`/`to_stdout` can be passed positionally, bypassing the guard. The suggested `inspect.signature(_convert_impl).bind_partial(...)` fix is correct. Worth a regression test (the current 16 are all keyword-only, so they wouldn't catch this or the case-sensitivity gap — `format="JSON"` slips through since the JAR lowercases but the wrapper doesn't).

**5. Could we split `batch_convert` into its own PR?** It's a new permanent public API and deserves its own design pass — separate from the #492 bugfix. A couple of things worth discussing there: the `Dict[path, None|Exception]` return reads a bit backwards (`None` = success), duplicate paths collapse dict keys, and since #451 is about *many* PDFs, a serial loop launching a fresh JVM per file is the slow path — `convert` already accepts a list/directory input plus `--threads`, so a single JAR invocation might be the better primitive.

Could you also switch `Closes #492`/`Closes #451` to `Refs` until the JAR-side fix lands? Happy to help with any of this — the direction is good, it mainly needs to land at the JAR layer.

PR #556, opened 2026-06-04
fix: preserve table header (TH) semantics in JSON output

### LINKED ISSUE #549: JSON output does not preserve table header cells (TH) — all cells serialized as TD, even with --use-struct-tree
## Summary

When extracting a **tagged PDF** whose table contains real header cells (`TH`), the **JSON output serializes every cell as `"pdfua_tag": "TD"`** and exposes no header indicator at all (`scope`, `is_header`, header type, …). The header information is clearly available internally — the **HTML output of the very same run emits correct `<th>` elements** — but it is dropped in the JSON serializer.

This makes the JSON output unsuitable for accessibility validation of tables (e.g. checking header presence/scope), even though the project is positioned to "Automate PDF accessibility".

## Environment

- opendataloader-pdf **2.4.7** (Python wrapper) / bundled JAR
- Java 21 (OpenJDK 21.0.11)
- Input: any tagged PDF whose table has `THead`/`TH` cells in its structure tree
- Flag: reproduced **both with and without** `--use-struct-tree`

## Steps to reproduce

```bash
# Same tagged PDF, same run
opendataloader-pdf -q -f json,html --use-struct-tree -o out tagged_table.pdf
```

The PDF's first table has 3 header cells (verified `THead > TR > TH` in the PDF structure tree).

## Actual JSON output (header info lost)

```json
{
  "type": "table",
  "pdfua_tag": "Table",
  "rows": [
    {
      "type": "table row",
      "cells": [
        { "type": "table cell", "pdfua_tag": "TD", "row number": 1, "column number": 1 },
        { "type": "table cell", "pdfua_tag": "TD", "row number": 1, "column number": 2 },
        { "type": "table cell", "pdfua_tag": "TD", "row number": 1, "column number": 3 }
      ]
    }
  ]
}
```

All cells are `TD`. Grepping the JSON for `"TH"`, `header`, `scope`, `is_header` returns nothing.

## HTML output of the same run (header info preserved ✅)

```html
<table border="1">
<tr>
  <th>...</th>
  <th></th>
  <th></th>
</tr>
</table>
```

So the engine *does* know these are header cells — only the JSON serializer omits it.

## Expected

The JSON should preserve the header semantics of the PDF structure tree, e.g. one of:

- set `"pdfua_tag": "TH"` for header cells (currently hardcoded/normalized to `TD`), and/or
- add an explicit field on the cell, e.g. `"is_header": true` and optionally `"scope": "row" | "col"` (mirroring the PDF `/Scope` attribute and HTML semantics).

## Why it matters

Accessibility tooling that consumes the JSON (table header presence/scope checks, PDF/UA validation) currently gets false negatives — a correctly tagged table looks header-less. Aligning the JSON with what the HTML output already does would close the gap.

## Notes

- Reproduced with `--use-struct-tree` and without; cell count differs between the two modes (the flag does change parsing), but in **both** the header flag is absent from JSON.
- The documented JSON schema (`schema.json`) for a table cell only includes `row number`, `column number`, `row span`, `column span`, `kids` — no header field.
- Happy to attach a minimal sample PDF if useful.

mvanhorn:
## Summary

Preserves table header semantics in the JSON output so the JSON and HTML paths agree on which cells are headers.

## Why

A tagged PDF table with real TH cells serialized every cell as `"pdfua_tag": "TD"` with no header indicator, even though the HTML output of the same run correctly emitted `<th>`. Issue #549 reports that this makes the JSON output unsuitable for table accessibility validation, because a consumer cannot tell header cells from body cells.

## Description

JSON output now preserves table header semantics so the JSON and HTML paths agree on which cells are headers. `TableCellSerializer.serialize()` emits an additive `is_header: true` field on first-row cells, mirroring the `isHeader = rowNumber == 0` heuristic `HtmlGenerator.writeTable()` already uses to emit `<th>`. The field name is a new `IS_HEADER` constant in `JsonName.java`, and the existing `pdfua_tag` field is left as-is so current consumers are not broken.

Before this change, a tagged PDF table with real TH cells serialized every cell as `"pdfua_tag": "TD"` with no header indicator, which made the JSON output unsuitable for table accessibility validation even though the HTML output of the same run emitted `<th>` (issue #549).

**Issue resolved by this Pull Request:**
Resolves #549

**Checklist:**

`LineArtSerializerTest.tableCellSerializerMarksFirstRowCellsAsHeaders` confirms first-row cells carry `is_header: true` and body cells omit the field. Covered by the new test in this PR; full suite runs in CI.

## Release Notes

* **New Features**
  * Table cells in the first row are now automatically identified and marked as headers in the generated output, improving data structure clarity.

* **Tests**
  * Added test coverage to validate header cell identification in the first row versus body cells.

### DIFF
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/json/JsonName.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/json/JsonName.java
@@ -50,6 +50,7 @@ public class JsonName {
     public static final String ROW_NUMBER = "row number";
     public static final String COLUMN_SPAN = "column span";
     public static final String ROW_SPAN = "row span";
+    public static final String IS_HEADER = "is_header";
     public static final String NUMBER_OF_ROWS = "number of rows";
     public static final String NUMBER_OF_COLUMNS = "number of columns";
     public static final String NUMBER_OF_PAGES = "number of pages";
--- a/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/json/serializers/TableCellSerializer.java
+++ b/java/opendataloader-pdf-core/src/main/java/org/opendataloader/pdf/json/serializers/TableCellSerializer.java
@@ -40,6 +40,9 @@ public class TableCellSerializer extends StdSerializer<TableBorderCell> {
         jsonGenerator.writeNumberField(JsonName.COLUMN_NUMBER, cell.getColNumber() + 1);
         jsonGenerator.writeNumberField(JsonName.ROW_SPAN, cell.getRowSpan());
         jsonGenerator.writeNumberField(JsonName.COLUMN_SPAN, cell.getColSpan());
+        if (cell.getRowNumber() == 0) {
+            jsonGenerator.writeBooleanField(JsonName.IS_HEADER, true);
+        }
         jsonGenerator.writeArrayFieldStart(JsonName.KIDS);
         for (IObject content : cell.getContents()) {
             if (!(content instanceof LineArtChunk)) {
--- a/java/opendataloader-pdf-core/src/test/java/org/opendataloader/pdf/json/serializers/LineArtSerializerTest.java
+++ b/java/opendataloader-pdf-core/src/test/java/org/opendataloader/pdf/json/serializers/LineArtSerializerTest.java
@@ -19,6 +19,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.jupiter.api.Test;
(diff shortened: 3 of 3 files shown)


MaximPlusov:
@mvanhorn Thank you for contribution! We added new method `isHeaderCell()` to `TableBorderCell`. Could you use this new method instead of `cell.getRowNumber() == 0`?

MaximPlusov (approved):

mvanhorn:
Glad to get this in, thanks @MaximPlusov. Preserving table header semantics in the JSON output keeps it aligned with the HTML path.

dennisangemi:
Thx <3

mvanhorn:
Thanks @MaximPlusov - preserving TH semantics so JSON and HTML agree on header cells is a clean consistency fix.

PR #569, opened 2026-06-12
fix(python): relay JAR stdout in quiet mode so --to-stdout and folder summary reach the caller

hyunhee-jo:
## Objective
Combining `--quiet` with `--to-stdout` produced empty output — pipe
consumers (`opendataloader-pdf --quiet --to-stdout f.pdf | jq ...`)
received nothing. The folder summary line (`Processed N PDF file(s) ...`)
was also swallowed under `--quiet`.

## Approach
In `run_jar` quiet mode, capture stderr (the JAR's log stream — that is
what quiet suppresses) but relay the captured stdout to the caller via
`sys.stdout.buffer`, reusing the encoding-safe pattern from streaming
mode (`errors="replace"`).

The JAR intentionally keeps payload and summary on stdout under `--quiet`
(CLIMain javadoc: the summary is a result, not a log); the wrapper
swallowing them was the bug, so this restores the documented contract
rather than changing it. The relay lives in `run_jar` rather than the CLI
entry point because `convert()` discards `run_jar`'s return value, so a
CLI-boundary relay would lose the summary for library callers.

### User-visible behavior change
Library callers using `convert(quiet=True)` (or `run_jar(..., quiet=True)`)
now see the JAR's stdout — typically one folder summary line — where they
previously got silence. Callers that `print()` the return value will see
the content twice; use either the return value or the relayed stream, not both.

## Evidence
Ran the packaged CLI on macOS the way a user would:

| Before: `--quiet --to-stdout --format text` | text on stdout | exit 0, stdout **0 bytes** |
| After: same command | full text on stdout | stdout **39,168 bytes**, stderr 0 bytes (quiet preserved) |
| After: folder run with `--quiet` | summary preserved | `Processed 1 PDF file in '...'` on stdout |
| After: failing run in quiet mode | streams surfaced once on stderr | unchanged — `check=True` raises before the relay |

Regression tests in `tests/test_runner.py` (4/4 pass):
- `test_quiet_success_relays_stdout_but_not_stderr` — payload relayed
  exactly once, stderr suppressed, return value intact.
- `test_quiet_relays_through_stdout_buffer_byte_path` — covers the
  production `sys.stdout.buffer` byte-relay path that `capsys` masks in
  the str-fallback test; a non-ASCII payload exercises the `utf-8` encode,
  and mutating the branch to str-only turns the test RED, proving it pins
  the byte path.

The wrapper suite's pre-existing docling import failures are unrelated and
identical before/after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Quiet mode now correctly relays the tool's standard output to the caller when using the "to-stdout" style flag, while continuing to suppress diagnostic error output.

* **Tests**
  * Added tests verifying quiet mode relays stdout but not stderr.
  * Added a test covering the byte-buffer relay path to ensure binary/stdout buffer forwarding works correctly.

### DIFF
--- a/python/opendataloader-pdf/src/opendataloader_pdf/runner.py
+++ b/python/opendataloader-pdf/src/opendataloader_pdf/runner.py
@@ -30,15 +30,27 @@ def run_jar(args: List[str], quiet: bool = False) -> str:
             ]
 
             if quiet:
-                # Quiet mode → capture all output
+                # Quiet mode → suppress the JAR's log stream (stderr) but
+                # relay its stdout to the caller: --to-stdout content and the
+                # folder summary line arrive on stdout, and swallowing them
+                # breaks pipe consumers (`... --quiet --to-stdout | jq`).
                 result = subprocess.run(
                     command,
-                    capture_output=True,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
                     text=True,
                     check=True,
                     encoding="utf-8",
                     errors="replace",
                 )
+                if result.stdout:
+                    if hasattr(sys.stdout, "buffer"):
+                        sys.stdout.buffer.write(
+                            result.stdout.encode("utf-8", errors="replace")
+                        )
+                        sys.stdout.buffer.flush()
+                    else:
+                        sys.stdout.write(result.stdout)
                 return result.stdout
 
             # Streaming mode → live output
--- a/python/opendataloader-pdf/tests/test_runner.py
+++ b/python/opendataloader-pdf/tests/test_runner.py
@@ -7,6 +7,7 @@ yet and is allowed to print the captured streams — but only once
 (``CalledProcessError.output`` and ``.stdout`` are the same attribute).
 """
 
+import io
 import subprocess
 from unittest.mock import MagicMock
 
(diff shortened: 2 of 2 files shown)


hnc-jglee (approved):

hnc-jglee (approved):

hnc-jglee:
## Multi-perspective review (dev / test / architecture / release)

**LGTM — approved for merge.** No changes requested.

Correctly diagnoses and fixes the regression where `--quiet --to-stdout` produced empty output. Root cause was that quiet mode captured stdout but then *discarded* it; the fix suppresses only stderr (the JAR log stream) while relaying stdout (payload + folder summary) to the caller — restoring the JAR's documented contract (`payload = result, stderr = log`).

All four perspectives agreed there is no blocking concern.

| Lens | Takeaway |
|---|---|
| Code quality | Relay, flush, and error paths are all correct. The str→bytes re-encode is unavoidable and lossless (U+FFFD is valid, `errors="replace"` is idempotent), matching the existing streaming-mode pattern. |
| Tests | Nicely done: the suite correctly identifies that `capsys` changes `sys.stdout.buffer` semantics (so the sibling test only exercises the str branch) and pins the production byte path separately via `io.BytesIO`. The RED→GREEN claim checks out. |
| Architecture | Behavior and scope are appropriate. Only a long-term cleanup note below. |
| Release | `npm run sync` is not needed here — this is a wrapper runtime-behavior change, not a JAR option change. |

### (FYI) Possible follow-ups — separate issues, unrelated to merging this PR
- **CLI/library API split**: applying the pattern Node.js already adopted in `60101c5` — `convert()` (no relay, returns) vs `_runForCli()` (relays, returns void) — would cleanly resolve `run_jar`'s dual "return + stdout side-effect" responsibility. Not something to do in this PR.
- **Real subprocess integration test**: tests are currently all mock-based, so actual encoding behavior and large-buffer handling are unverified. A fake-jar script would strengthen this.

Both are nice-to-haves and do not block the merge. 👍

<sub>🤖 4-perspective agent team review</sub>

hyunhee-jo:
Thanks for the thorough review. On the follow-ups:

- **CLI/library API split** — filed as #572 to track porting the Node #473 pattern. Kept out of this PR since it pairs a behavior change with generated-file regeneration.
- **Real subprocess integration test** — agreed it's a nice-to-have, but leaving it unfiled for now given the low priority; can revisit if the mock-based coverage proves insufficient.

Issue #686, opened 2026-08-17
opendataloarder gets stuck


abhaysap:
### Bug

opendataloarder gets stuck  when we try to convert pdf to md in hybrid mode 

### Steps to reproduce

on server 

bash: /home/sunny/perl5/perlbrew/etc/bashrc: No such file or directory
sunny@sunny-XPS-15-9510:~$ source ~/venvs/opendataloader/bin/activate
(opendataloader) sunny@sunny-XPS-15-9510:~$ opendataloader-pdf-hybrid --port 5002  --ocr-engine auto --enrich-formula --enrich-picture-description
2026-08-17 17:06:13,749 - INFO - Loading plugin 'docling_defaults'
2026-08-17 17:06:13,758 - INFO - Registered ocr engines: ['auto', 'easyocr', 'kserve_v2_ocr', 'nemotron-ocr', 'ocrmac', 'rapidocr', 'tesserocr', 'tesseract']
2026-08-17 17:06:14,317 - INFO - Accelerator: CPU (no GPU detected)
2026-08-17 17:06:14,317 - INFO - Starting Docling Fast Server on http://0.0.0.0:5002
2026-08-17 17:06:14,317 - INFO - OCR settings: do_ocr=True, ocr_engine=auto, force_ocr=False, lang=default
2026-08-17 17:06:14,317 - INFO - Max file size: unlimited
2026-08-17 17:06:14,317 - INFO - Enrichments enabled: formula, picture-description
INFO:     Started server process [452660]
INFO:     Waiting for application startup.
2026-08-17 17:06:14,348 - INFO - Initializing DocumentConverter (do_ocr=True, ocr_engine=auto, force_ocr=False, lang=default, enrichments=formula,picture-description, device=auto)...
2026-08-17 17:06:16,227 - INFO - DocumentConverter initialized in 1.88s
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:5002 (Press CTRL+C to quit)
INFO:     127.0.0.1:50250 - "GET /health HTTP/1.1" 200 OK
2026-08-17 17:11:32,914 - INFO - detected formats: [<InputFormat.PDF: 'pdf'>]
2026-08-17 17:11:32,992 - INFO - Going to convert document batch...
2026-08-17 17:11:32,993 - INFO - Initializing pipeline for StandardPdfPipeline with options hash 7060884fc20f485e6edab632a836312e
[...]

PR #752, opened 2026-09-23
feat(hybrid): let the operator bound page processing with --document-timeout

### LINKED ISSUE #749: Hybrid server has no upper bound on a single conversion
## Problem

The hybrid server sets no upper bound on a single conversion:

- `PdfPipelineOptions` is built without `document_timeout`, so docling's default of `None` applies — no limit.
- `uvicorn.run()` is started without a request timeout.
- Conversions are serialized behind `_convert_lock`, so one long conversion blocks every other client for its whole duration.
- `--max-file-size` defaults to `0` (unlimited), and file size is a poor proxy for conversion cost anyway: a small PDF can hold thousands of images or pathological content streams.

## Impact

A single upload can occupy the server indefinitely, and the only trace is the `queued` line other requests emit while they wait. This matters most with `--enrich-picture-description`, where cost follows picture count and every picture is described by default (see #747): a small file holding many images turns into hours of VLM inference under one request.

The module documents itself as having no authentication and binds `0.0.0.0`, so on a network where callers are not trusted, the work a caller can request has no ceiling.

## Fix

Set `document_timeout` from a new CLI flag with a sane default, so a conversion that exceeds it fails as a partial result rather than running forever. docling reports a timeout as `PARTIAL_SUCCESS`, which this server already surfaces through `failed_pages`, so the response shape needs no change.

Worth deciding alongside it: whether the default should bound wall-clock time per document, per page, or both.

bundolee:
Objective: a single conversion had no upper bound (#749). docling supports one,
this server never set it, and conversions are serialized behind one lock -- so
one pathological upload holds up every other client for as long as it runs,
which is what "opendataloarder gets stuck" (#686) looks like from the outside.

Approach: expose docling's `document_timeout` as --document-timeout, in seconds,
with the default left unlimited so nothing changes for a workstation converting
its own files. 0 spells "no limit"; handing that to docling would instead abort
every conversion after its first page batch. Non-finite values are rejected,
since `inf` would be reported at startup as a limit while enforcing none.

A timeout also has to be legible downstream, and it was not. docling names the
abandoned page in `ErrorItem.page_no` and leaves that page's entry in `pages`,
so both existing failed-page strategies -- message parsing and gap detection --
come back empty. This server kept only the error messages and dropped `page_no`,
so the client read a truncated document as a whole one and wrote blank pages
with no warning. Those pages now reach `failed_pages`, which is what makes the
loss visible.

Evidence: 6-page paper at --document-timeout 1, before and after the page_no fix.

| Stage                    | failed_pages | Java CLI                                   |
|--------------------------|--------------|--------------------------------------------|
| before                   | `[]`         | no fallback, blank pages, exit 0           |
| after                    | `[1..6]`     | "fallback disabled -- failing fast", exit 1 |
| after, --hybrid-fallback | `[1..6]`     | reprocesses them locally, 173 items, exit 0 |

On an 837-page sample at --document-timeout 20: 42s wall for a 20s limit, 5 of
837 pages converted, and before the fix the CLI wrote 180 typed items for the
whole book and exited 0.

What the limit does not cover is stated in README and --help rather than left to
be discovered: docling checks the clock between page batches, so a conversion
overruns the number by whatever is in flight (~2x measured), and the enrichment
stages are not checked at all -- a document whose cost sits in picture
description can run far past the limit and still come back a plain success.

Fixes [opendataloader-project/opendataloader-pdf#749](https://github.com/opendataloader-project/opendataloader-pdf/issues/749)

## Review

Reviewed from three angles before asking for eyes on it: correctness, security, and comment
policy. All three failed the first draft, and two of them independently found the same defect
— the timeout was unobservable downstream, so enabling the flag introduced silent truncation
that had not existed. That is the `page_no` half of this PR; the first draft only set the
docling option.

The same reviews established what this flag cannot do, which is now documented rather than
implied. A real ceiling on conversion cost needs the enrichment stages covered too; that is
docling's to fix, and an upstream issue is worth filing.

Mutation-checked the tests: dropping the docling option, the `or None`, the `create_app`
wiring, the production validator's `type=`, the `page_no` collection, or the timeout cause in
the log each fails at least one test. The `create_app` wiring and the validator were both
unprotected in the first draft.

## Follow-ups

- docling does not check `document_timeout` during enrichment (`base_pipeline._enrich_document`),
  so `--enrich-picture-description` is unbounded. Measured 49x and 86x overruns on a 10s limit.
- `/v1/profile/file` converts one upload three times, one profile with picture description
  forced on. #747 adds the `--max-file-size` check it was missing.

## Summary by CodeRabbit

* **New Features**
  * Added an optional `--document-timeout` setting to limit the time spent converting a document.
  * When a timeout is exceeded, conversions return partial results and identify abandoned pages as failed, allowing clients to fall back on them.
  * Timeout events are identified in server logs. No timeout applies by default, and negative or non-finite values are rejected.
* **Documentation**
  * Added guidance on timeout behavior, including when limits are checked and how to pair them with file-size limits.


bundolee:
Closing this: `--document-timeout` cannot be offered as a bound on conversion cost.

docling checks `document_timeout` only in the page-processing loop (`standard_pdf_pipeline._build_document`). `_enrich_document` has no check, and `execute()` runs it unconditionally after a build timeout — so with `--enrich-picture-description` on, the limit does nothing. Measured against a 10s limit: 6 pages with 6 pictures ran 498s and returned `SUCCESS` with `has_timeout_errors() == False`; 12 pages with 11 pictures ran 865s. Even for page processing alone the limit is checked between batches, so a conversion overruns it by whatever is in flight — about 2x in local runs.

A flag that names a number it does not enforce is worse than no flag, so this is withdrawn rather than shipped with caveats.

One finding from reviewing it stands on its own and is now #754: docling reports a failed page in `ErrorItem.page_no` while leaving that page's entry in `pages`, and this server dropped `page_no`, so `failed_pages` came back empty and the Java client wrote blank pages with exit 0. That is independent of timeouts — docling sets `page_no` for BACKEND_FAILURE and UNKNOWN failures too.

The branch is left in place for reference.

PR #753, opened 2026-09-23
ci: land the version bump through a pull request

bundolee:
The v2.5.11 release published every artifact and then failed to move `main` off
the released coordinate: three `git push` attempts, three `GH013 — Changes must
be made through a pull request`. The PAT added for v2.5.10 did not help, because
the ruleset's bypass list holds only the organization-admin and repository-admin
roles and that token's owner holds neither. `github-actions` cannot be listed at
all — it is a runtime actor, not an account.

So this stops pushing past the rule and satisfies it. The `default` ruleset asks
for `required_approving_review_count: 0`, sets `require_code_owner_review: false`
and declares no required status checks, and permits `rebase` as a merge method.
A pull request opened by the release run therefore has nothing left to wait for
and can be rebase-merged in the same step.

## Changes

- **`scripts/open-version-bump-pr.sh`** (new) — builds the bump on `main` in a
  worktree under `RUNNER_TEMP`, pushes the branch, opens the PR and rebase-merges
  it, retrying the whole sequence up to three times if `main` moves underneath.
  Because the tag's tree is never checked out of the way, the `trap` that used to
  restore it for the docs steps is gone.
- **Token** — the step now runs on the run's own `GITHUB_TOKEN` with
  `pull-requests: write`. `HOMEPAGE_SYNC_TOKEN` is left with one job, the
  reference-docs push, and `scripts/preflight.sh` no longer checks it against
  this repository.
- **`snapshot.yml` dispatch** — loop prevention silences the push that a
  `GITHUB_TOKEN` merge produces, so the step asks for the run explicitly
  (`actions: write`) instead of relying on the merge to trigger it.
- **Step order** — the bump is now the job's last step. As step 19 of 21 its
  failure skipped the two docs steps behind it, which is why the homepage
  reference docs are still missing `--hybrid-chunk-size`.

## Verification

`scripts/set-dev-version.sh` was run in a clean worktree on this machine to
confirm `mvn versions:set` and `pnpm version` both work without `node_modules`
or a prior build; it produced exactly the eight expected files. Both workflows
parse, both scripts pass `bash -n`, and the ruleset parameters quoted above are
from `GET /repos/:owner/:repo/rulesets/7000669`.

One assumption is not verifiable ahead of a real release: the ruleset also sets
`require_extra_approval_for_unattributed_changes: true`, which would raise the
required approvals from 0 to 1 for a commit GitHub cannot attribute to an
account. The bump commits as `github-actions[bot]@users.noreply.github.com`,
which GitHub does attribute to that bot, so this is expected to pass — but the
first release after this merge is the proof. If it does fail, the step's error
output names the reason, and the fix is either a bypass entry or turning that
sub-rule off.

## Summary by CodeRabbit

* **Chores**
  * Release version updates are now submitted through a pull request instead of being pushed directly to the main branch.
  * Snapshot builds are dispatched after the version update is merged, or when the main branch already has the target version.
  * Documentation synchronization now runs as a separate step in the release process.


bundolee:
## Review round — three passes, applied in 513cf8bc

Reviewed for correctness, security and comment discipline. One finding was a vulnerability this PR itself introduced.

### Critical — fork pull request hijack

`gh pr list --head` filters on the ref name and nothing else; its own help says `Filter by head branch ("<owner>:<branch>" syntax not supported)`. The bump's branch name follows from the version main declares, so it is known in advance.

Anyone able to open a pull request — no write access needed — could push `chore/begin-X.Y.Z-snapshot` to a fork, open it against `main`, and wait. The release run's lookup would find that pull request, skip `gh pr create` because one for the head already appeared to exist, and merge it with `--rebase`. Zero reviews, because the ruleset asks for zero. The `gh workflow run snapshot.yml` call right after would then run the merged code in a job holding the Maven Central credentials.

The lookup now selects on `isCrossRepository == false` and `headRefOid == $(git rev-parse HEAD)`, and the merge pins that same commit with `--match-head-commit`, which also closes the window between reading mergeability and merging.

### Correctness

- The retry loop guarded two of the eight commands that could fail; everything else died on `set -e`, including a transient `gh pr view` during the mergeability poll. The attempt is now a single guarded function, and a merge refused off a stale mergeability answer is retried instead of failing the attempt.
- A run that gave up left the branch and its pull request open indefinitely — a hand merge weeks later would have applied a bump built on a long-gone main. Both are discarded now.
- `git add --update` replaces the fixed eight-path list, which could disagree with the tree-wide `git diff --quiet` above it as soon as a module is added.
- The early "main already declares it" exit skipped the snapshot dispatch and left earlier attempts' residue behind.

### Structure

The bump is now its own job beside the docs sync, both on `needs: release`. As consecutive steps, whichever ran first could skip the other by failing — the previous commit fixed that in one direction and opened it in the other, since a homepage sync failure would now block the bump. As siblings, neither can. The bump job also no longer sits next to `id-token: write`, the GPG key or the registry tokens, so its `actions: write` and `pull-requests: write` are out of reach of the deploy steps' third-party actions, only one of which is SHA-pinned. Its checkout is a clean tag tree, so the worktree the script needed is gone.

### Withdrawn from the original description

`require_extra_approval_for_unattributed_changes: true` does not apply here: the setting has no effect when a ruleset requires zero approvals. The assumption flagged as unverifiable is moot, and it would not have blocked the fork attack either.

### Verified

- `can_approve_pull_request_reviews: true` on this repository, so Actions may open the pull request. The organization-level value needs `admin:org` and was not readable; a create refused for that reason now says so.
- Retry control flow exercised across five outcome sequences, the head-filter jq selector across four (ours / fork with matching SHA / same repo with stale SHA / empty).
- `set-dev-version.sh` run in a clean worktree: eight files, nothing else tracked touched, so `--update` stages exactly what the check saw.
- Both workflows parse; both scripts pass `bash -n` and `shellcheck` (one SC2329 false positive on a trap-invoked function).

### Not changed

- `git push --force` on the bump branch would overwrite a human branch of the same name. The name belongs to the release machinery.
- `set-dev-version.sh` and the poms come from `main` while the orchestration comes from the tag. Pre-existing, and reaching it needs merge access to main.
- `HOMEPAGE_SYNC_TOKEN` may still hold write access to this repository although nothing here uses it any more. Worth reissuing against the homepage repository alone — separate from this PR.

